PR #28, opened 2021-09-23
Topic/kdhaduk/beta 1

dhadukk:
Kapp-controller related changes and fixes based on some of the feedback


jncd:
Hi @dhadukk 
Thanks for making this PR to the beta1 docs too.
What does it mean to use a "suitable" version? How does a customer know which versions are suitable and which ones are not?

dhadukk:
Hi @jncd 
I meant user could choose to install kapp -controller version greater than v0.20.0. Maybe you could help in rephrasing it. Thanks.

PR #29, opened 2021-09-23
Add Service Account Error to the doc

vibhas:
Hi @jncd, I have added the Service Account Error to the troubleshooting doc. I did not add the third error about the timeout. I think it might be worth waiting for some more evidence from the users before we add that.


jncd:
Thanks @vibhas

PR #31, opened 2021-09-24
Instructions to install Supply Chain Security Tools - Store

Manifaust:
Here are instructions to install the metadata store. I copied the instructions from CNR and customized it.

I'm looking for some feedback. I notice that the `values.yml` for the other pacakges (like CNR) all have a section for registry, something like this:
```yaml
    registry:
      server: "registry.pivotal.io"
      username: "TANZU-NET-USER"
      password: "TANZU-NET-PASSWORD"
```

Are these configuration settings supported by default by the Tanzu CLI install feature, or did each package author implement support for these values?

Second, I didn't list out all the possible configuration in our example `values.yaml`, should I? In our project README, we link to another doc where we describe each property in more detail, I'm not sure what we should do here.

Thanks for your input.

### DIFF
--- a/install.md
+++ b/install.md
@@ -668,6 +668,82 @@ To install Application Live View:
     see the [Application Live View documentation](https://docs.vmware.com/en/Application-Live-View-for-VMware-Tanzu/0.1/docs/GUID-index.html).
 
 
+### <a id='install-scst-store'></a> Install Supply Chain Security Tools - Store
+
+To install SCST - Store:
+
+1. Follow the instructions in [Install Packages](#install-packages) above.
+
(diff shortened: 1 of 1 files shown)


jncd:
Thanks @Manifaust

jncd:
Hi @Manifaust Tony!
I've got your PR up on staging:
https://docs-staging.vmware.com/en/VMware-Tanzu-Application-Platform/0.2/tap-0-2/GUID-install.html#install-supply-chain-security-tools--store-13

We had to do a little work on formatting/indenting, make sure it's all still good.

PR #34, opened 2021-09-24
Update Accelerator install instructions

trisberg:
Samples are now included in the package so no extra step needed


jncd:
Thanks @trisberg

jncd:
This is the best fix! Thanks @trisberg

PR #64, opened 2021-09-28
Add docs for API portal

fombico:
You can see the sample formatting on my repo: 
https://github.com/fombico/docs-tap/blob/main/install.md#-install-api-portal

### DIFF
--- a/install.md
+++ b/install.md
@@ -120,6 +120,7 @@ To install any package from the Tanzu Application Platform package repository:
     + [Install Supply Chain Security Tools - Store](#install-scst-store)
     + [Install Supply Chain Security Tools - Sign](#install-scst-sign)
     + [Install Supply Chain Security Tools - Scan](#install-scst-scan)
+    + [Install API portal](#install-api-portal)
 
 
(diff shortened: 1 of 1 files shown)


jncd:
Hello @fombico 
Your docs are merged and on the staging site. w00t!  Take a look https://docs-staging.vmware.com/en/VMware-Tanzu-Application-Platform/0.2/tap-0-2/GUID-install.html#install-api-portal-9
I suspect when you see it, you'll want to do a little more of that "pattern matching" that you engineers are so good at. 

Suggestions:
+ Use numbered steps to contain everything that the user has to do. (You can probably turn each heading into a step.)
+ Use ALL-CAPS for placeholder variables so `{version}` => `VERSION`
+ Use title case, Chicago, for all headings. [Capitalize my title](https://capitalizemytitle.com/) So Install API portal" is "Install API Portal".
+ Here's my best guide to formatting cross-references: https://docs-wiki.sc2-04-pcf1-apps.oc.vmware.com/wiki/style/cross-ref-style.html

Back in the olden days, you might have had a tech writer to help with this, but VMware appears to have stopped hiring them. 🤷🏽

PR #68, opened 2021-09-29
Add instructions to create cluster image policy

derwei:
Since the cluster image policy is a necessary step for the proper functioning of our webhook, we are adding these instructions to create the cluster image policy to the install docs. We currently have it in the getting started docs.


jncd:
Thanks @derwei

PR #71, opened 2021-09-29
Update install-general.md

mpriamo:
- removed TKG references
- Removed steps prereq and cluster config 
- fixed headers


jncd:
Thanks @mpriamo

PR #73, opened 2021-09-29
Update API portal docs to use imperative language

fombico:
Regarding the app title - it is actually called "API portal" (where 'portal' is in lowercase). 
My PM told me to lowercase it in our code and other documentation.


jncd:
Thanks!! @fombico 

Really good to know about the product name. I'll update the other places in the docs where there is a capital "P".

PR #83, opened 2021-09-30
Add accelerator plugin install instructions

zrob:


jncd:
Hi @zrob 
We had some merge conflicts with this.
Someone decided that the plugin install needs to be broken out separately.

a-peek4:
@zrob I kept the content you added about installing the app accelerator plugin to the ## Install the Tanzu CLI Plugins section at the bottom of the doc. If we still need windows and mac specific instructions for installing the plugins, please create a new PR with those. However, it looked like the instructions you added under ## Install the Tanzu CLI Plugins may work for all operating systems?

PR #88, opened 2021-09-30
remove reference to package plugin (step 4)

heyjcollins:
the package plugin comes pre-installed so shouldn't be referenced directly in this copy.
there are other plugins that must be installed, so referring to them generally is appropriate.


jncd:
Thanks @heyjcollins

PR #90, opened 2021-09-30
Redecoration on text style

N00BEST:
Signed-off-by: Daniel Marcano <dmarcano@vmware.com>


jncd:
Thanks @N00BEST

jncd:
I'm going to post this to staging now so you can see the improvement in the lists.

Next top tip: Less of the "we" if you can. I know it's an occupational hazard because you did the coding, but the reader is more interested in what _they_ can do and what is there now for _them_. (They are less interested in what you did — sad but true.)

PR #93, opened 2021-10-01
add troubleshooting steps for mac CLI install

heyjcollins:
Mac prevents users from running software downloaded via a browser.
I've added steps to follow to work around this limitation.

I also updated a typo in the verification steps for installing cert-manager


jncd:
Thanks @heyjcollins

PR #95, opened 2021-10-01
add all packages returned by `package list` command

heyjcollins:
the example output only had a few packages so I've updated to reflect the full set of packages that a user can expect to see in the list for beta2


jncd:
Thanks @heyjcollins

PR #97, opened 2021-10-01
update cnr version , fix links

heyjcollins:
version of cnr is 1.0.2 now so updated it.
also fixed some inline links


jncd:
Thanks @heyjcollins

PR #99, opened 2021-10-01
re-order package installs based on dependencies

heyjcollins:
supply chains depend on a few pre-requisite packages.
app live view does as well...

so moved those packages above so they'll already be installed when the user gets to installing supply chains and app live view


jncd:
thanks @heyjcollins

PR #109, opened 2021-10-01
simlify applying k8s resources

ryan-a-baker:
Allow creating k8s resources without having to put them in a file then `kubectl apply -f ...`.  I left this out of any of the ones where you have to replace a variable with a value pulled from the environment, but could apply those if desired. 

 I probably would make a single command to pull the value then apply it, which would take a bit longer to test out, but would like to get the others in before bug bash.


jncd:
Thanks @ryan-a-baker !

PR #114, opened 2021-10-01
Update install instruction to warn of known issue

Manifaust:
SCST - Store has a known issue where we do not allow users to the change database passwords between deployment. I warn the user in install instructions and link to the known issue doc.


jncd:
Keep 'em coming. Thanks @Manifaust

PR #117, opened 2021-10-01
Clarify SCP naming

teddyking:
Small PR to spell out what SCP mean (Services Control Plane).


jncd:
Thanks @teddyking 
Make sure with the PMs that you can abbreviate the SCP name at all. We're usually not allowed to do it. (I expect that eventually, we'll have to spell it out in full everywhere unless it's actually in a code sample.)

PR #120, opened 2021-10-01
Update TBS Install Command

mgibson1121:


jncd:
Thanks @mgibson1121 

@heyjcollins I'm merging this even without your review (stuff is changing so fast that I want to avoid merge conflicts). Please review for @mgibson1121 and make another little PR with any corrections.

PR #122, opened 2021-10-01
Add more links to TOC

Manifaust:


jncd:
Thanks @Manifaust 

Yes, you need a blank line after the stem sentence introducing a bulleted or numbered list. (Not needed in regular markdown but somehow required in the VMware publishing system.)

Top tech writing tip: Spell out all digits from zero to nine. (So, "Two components" rather than "2 components". It does make a subliminal difference.)

PR #123, opened 2021-10-01
Update docs around updating Tanzu CLI

alexbarbato:
NOTE: There is a long term good recommendation we left behind around updating the CLI. Not sure if that's the appropriate way to handle this.


jncd:
Thanks @alexbarbato 

We'd noticed the angle brackets but hadn't had time to investiate. (@paigecalvert )

PR #124, opened 2021-10-01
Fix code blocks for easy copy/pasting

derwei:
Separated out commands from output


jncd:
Thanks @derwei 

PR #125, opened 2021-10-01
Fix TOC not supporting bullets that aren't also links

Manifaust:


jncd:
Thanks @Manifaust 
These TOCs are tricky.

PR #127, opened 2021-10-01
Use more consistent titles for Scan and Store

Manifaust:


jncd:
THANK YOU

@Manifaust

PR #128, opened 2021-10-01
Updated flux crd install section

rashedkvm:
Signed-off-by: Rashed Kamal <krashed@vmware.com>


jncd:
Thanks @rashedkvm 

btw: what's there now looks more like two steps in sequence (a little procedure) that should be turned into a numbered list with a stem sentence to introduce it.

PR #130, opened 2021-10-01
Update EULA product lists

C10U:
Update TAP Beta2 products which EULA needs to signed separately to reflect the improved UX


jncd:
Thanks @C10U 

Nice one! 

Actually, I was thrilled to see on the product names list that an allowed abbreviated form is "API portal" — like you not having to add "for Tanzu" everywhere. 

General rule is to use the full product name on first use in body text on the page. After that (and in headings) the abbreviated form is allowed.

PR #133, opened 2021-10-01
Add app accelerator UI instructions link

trisberg:


jncd:
Thanks @trisberg 

Top tip: Don't use "please" in customer-facing documentation, except when you're asking the customer to do VMware a favour. (As in "Please send us feedback." [so we can improve the product and make more money].

But, if you're just asking them to do themselves the favour — like see this doc so you can do your job. We skip the "please". Harsh? Maybe. But less is more.

PR #135, opened 2021-10-01
Updating TBS install to use poll-timeout

alexbarbato:
`--poll-duration` doesn't appear to be a CLI option.


jncd:
Thanks @alexbarbato 

It's all in the details 😉

PR #137, opened 2021-10-01
Add namespace to commands;Add command to create secret

kkavitha:


jncd:
Thanks @kkavitha 

For placeholders (variables) in docs, we always use capital letters. So, `secret-1` will have to be `SECRET-1`. Will you do the change for us?

PR #139, opened 2021-10-01
adding convention service explanation

heyjcollins:
to explain more about the two separate components which make up the service

### DIFF
--- a/install.md
+++ b/install.md
@@ -323,9 +323,17 @@ They are only used for configuration of non-standard installs.
 
 ## <a id='install-convention-service'></a> Install Convention Service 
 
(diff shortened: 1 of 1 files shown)


jncd:
Thanks @heyjcollins 

The PR rate has slowed a little so I had a go at editing in situ.

Top tip for tech writing: 
+ Keep lines short, ~100 - 120 chars. (Lots of advantages including fewer merge conflicts.) 
+ Use present tense unless you are talking of a specific time in the moderate to distant future. "Convention Services _are installed_ separately" unless you mean "Convention Services _will be installed_ separately in a future release."

heyjcollins:
thanks for the guidance and for all the mergings @jncd !

PR #142, opened 2021-10-01
Clean-up documentation for creating a service account for public vs private registries

stormqueen1990:


jncd:
Thanks @stormqueen1990

PR #143, opened 2021-10-01
Fix source-test-to-url supplychain instructions

sclevine:
The section for the source-test-to-url SupplyChain is currently a duplicate of the previous section.


jncd:
Thanks @sclevine 

Looks good to me!

PR #145, opened 2021-10-01
Update secret instruction for default supply chain

kiwi-bui:


install.md
```diff
@@ -466,17 +466,18 @@ You install Convention Servers are part of separate installation procedures. For
...
-      repository: ${registry_repository}")
+      server: REGISTRY_SERVER
+      repository: REGISTRY_REPOSITORY
     ```
 
 2. Create a imagepullsecret:
 
     ```
-    tanzu imagepullsecret add tanzunet-credentials --registry registry.tanzu.vmware.com --username TANZU-NET-USER --password TANZU-NET-PASSWORD --export-to-all-namespaces || true
+    tanzu imagepullsecret add registry-credentials --registry <REGISTRY> --username <REGISTRY_USERNAME> --password <REGISTRY_PASSWORD> --export-to-all-namespaces || true
```
paulcwarren, line 478: Maybe REGISTRY should be REGISTRY_SERVER?

jncd:
Hello @kiwi-bui 
We're merging this but would you do another check on it:
For placeholders, we always use uppercase with hyphens and never angle brackets.

jncd:
If `REGISTRY_USERNAME` is an environment variable that gets typed as is then we'd use the underscores. But otherwise, it's hyphens. It's a neat way to keep the two different.

PR #147, opened 2021-10-01
Update description of inner loop

zrob:


paigecalvert:
thank you @zrob! I'm glad you all spent the time to edit this description. We'll push the staging site soon, so you can check the formatting etc. We'll also do an editing pass for style, language etc.  

qq: Could you clarify the differences between:

"writing code, committing to a version control system, testing, and then making additional code changes."
and 
"coding, testing, and debugging until they are happy with their code changes and finally make a commit. Rinse and repeat."

Is the important piece here that testing and debugging happens before they make the first commit?

zrob:
hey @paigecalvert 

we talk about inner loop being the things you do BEFORE committing. once a commit happens, then those changes move on to outer loop activity.  this usually means automated deployment to integration/staging environments.

the big change here was to remove the deployment piece and staging reference  from the inner loop description.

and the more nuanced change was to position commit at the end of inner loop.  the commit isn't repetitive for a change. it's the last part of the change.  the code/test/debug loop are the main repetitive pieces of the loop.

hope that helps

PR #148, opened 2021-10-01
Update install-general.md

mingxiao:
Add --yes to kapp deploy commands


jncd:
Awesome!
Thanks @mingxiao 

How're you doing?

PR #150, opened 2021-10-01
Add GKE as supported

evankanderson:
Per discussion with Valentina in Slack.

### DIFF
--- a/install-general.md
+++ b/install-general.md
@@ -16,6 +16,7 @@ The following prerequisites are required to install Tanzu Application Platform:
 
     * Azure Kubernetes Service
     * Amazon Elastic Kubernetes Service
+    * Google Kubernetes Engine (note that GKE Autopilot clusters do not have required features enabled)
     * kind
     * minikube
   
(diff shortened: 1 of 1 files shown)


jncd:
Thanks @evankanderson 

I can merge this. 
But, maybe you can clarify the note... What does it mean?
How does the reader learn what "required features" they have to enable on their autopilot clusters? 
Or, do you mean that I can't use autopilot clusters but there are other types of GKE clusters can be used for TAP?

evankanderson:
GKE Autopilot clusters historically haven't supported [mutating admission controllers] (https://cloud.google.com/kubernetes-engine/docs/concepts/autopilot-overview#webhooks_limitations), though it looks like GKE 1.21 may reduce that limitation. It's not clear that we've written all our webhooks to respect that, though 

The default configuration of GKE clusters should work fine.

a-peek4:
@evankanderson can you specifically list which required features are not enabled? Is the take away here that GKE is supported, EXCEPT for GKE auto pilot clusters?

Can users take steps to enable required features for GKE autopilot clusters?

This is merged into the docs, however I see this causing confusion for users if we cannot provide more detailed info.

evankanderson:
GKE Autopilot does not support mutating admission webhooks, which are a core Kubernetes API used for defaulting values. There is no way to enable this for the Autopilot version of GKE; regular GKE clusters should work fine.

evankanderson:
I'm fine if we want to reword this or put it in a footnote; my concern was having people put a lot of effort into debugging a scenario where the cluster couldn't be fixed.

PR #151, opened 2021-10-01
Improve CNR install documentation

evankanderson:
Mention local Kind port configuration and explain provider: local


jncd:
Hi @evankanderson 
I fixed the merge conflict but still wonder what I should do with: 
> if using Cloud Native Runtimes, you may want to [configure kind according to this guide](https://docs.vmware.com/en/Cloud-Native-Runtimes-for-VMware-Tanzu/1.0/tanzu-cloud-native-runtimes-1-0/GUID-local-dns.html#configure-your-local-kind-cluster-1))

_may want to_ is pretty vague. I do know if I'm allowed to want this? If I don't want it, will everything be just fine? 

Or could we state something more direct? 
> If you are using Cloud Native Runtimes, see [Configure Your Local Kind Cluster](https://docs.vmware.com/en/Cloud-Native-Runtimes-for-VMware-Tanzu/1.0/tanzu-cloud-native-runtimes-1-0/GUID-local-dns.html#configure-your-local-kind-cluster-1) in the Cloud Native Runtimes documentation to configure kind.

evankanderson:
I'm fine with that wording; the install without those directions will work, but it gets harder to access services on your cluster later.

jncd:
Thanks ! I made the change.
Screen Shot 2021-10-01 at 5 54 14 PM

evankanderson:
(and you can't change the kind options after cluster create)

PR #153, opened 2021-10-02
Fixing the update command in the troubleshooting doc

vibhas:


jncd:
Hi @vibhas 
Thanks for the fix. Sad to see you working on the weekend.

PR #154, opened 2021-10-04
Move kubectl prereq towards the top

KeithRichardLee:
Move kubectl prereq towards the top as following instructions require kubectl


jncd:
Thanks @KeithRichardLee

PR #156, opened 2021-10-04
remove ".com" from kp_default_repository

KeithRichardLee:
remove ".com" from kp_default_repository as technically if user replaced EXAMPLE-REGISTRY with their reg url it means there be a .com on the end!


jncd:
Thanks @KeithRichardLee

PR #167, opened 2021-10-04
remove `-dev` from tanzu-cli folder downloads

heyjcollins:


jncd:
Sweet! Thanks @heyjcollins 

What about this one? Surely it too? I'm going to make it go away as part of this story. 

Screen Shot 2021-10-04 at 12 19 25 PM

PR #169, opened 2021-10-04
Fix formating and commands to be more consistent, add prerequsites se…

derwei:
…ction at the start so users will have everything they need to go through the install.


jncd:
@derwei 
Could you take a look at these merge conflicts?

derwei:
Hi @jncd ! I've fixed the conflicts. Please let me know if there's anything else I can help with or need to address.

jncd:
Thanks so much @derwei 

Top tip: I think this isn't going to format correctly in your output. 

For VMware markdown (but not for github markdown), you need a blank line between stem sentence (with its colon) and the start of bullets or numbered steps. 

I'll push and we'll see.

Screen Shot 2021-10-04 at 2 19 09 PM

derwei:
yup! you're right! looks like that broke things. let me push a fix

PR #173, opened 2021-10-04
Doc update - clarify dependency

lisamburns:


jncd:
Thanks @lisamburns 
Looks good!

PR #174, opened 2021-10-05
Removing unnecessary default supply chain steps

DanieldeR:
- Removing duplicated entry in the TOC for default supply chains
- Remove unnecessary steps in the installation of the default supply chains


jncd:
Thanks @DanieldeR 

So, it's supposed to be "Default Supply Chain" singular...  Good to know. I'll add that to the readme.

PR #175, opened 2021-10-05
Update CNR installation guide

tanzeeb:
Revamp steps 1 and 2 for the CNR installation:

* Remove references to TKG, speak about "existing Contour installations" generically
* Un-merge step 1  into two steps (install packages, gather values schema). It flows a lot better now that the schema values have detailed descriptions and defaults.
* Clearly outline the two special conditions (local cluster, existing contour) for specifying `cnr-values.yaml`, and note that the defaults are fine for all other situations

Incorporates feedback from https://jira.eng.vmware.com/browse/TANZUSC-344 and others.

### DIFF
--- a/install.md
+++ b/install.md
@@ -174,10 +174,8 @@ To install any package from the Tanzu Application Platform package repository:
 To install Cloud Native Runtimes:
 
 1. Follow the instructions in [Install Packages](#install-packages) above.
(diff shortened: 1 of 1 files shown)


jncd:
Thanks @tanzeeb 
Looks pretty good.

PR #176, opened 2021-10-05
Add additional pre-reqs, fix other pre-reqs

sclevine:
Best effort to address TANZUSC-322 and TANZUSC-328.


jncd:
Thanks @sclevine

PR #178, opened 2021-10-05
Make prereq versions consistent

sclevine:


jncd:
Thanks @sclevine 
When is source controller two words and when is it hyphenated?

sclevine:
`source-controller` is the name of the repository/component, but it's such a generic name that it needs to be prefixed with Flux for clarity.

PR #180, opened 2021-10-05
Add section for Workload namespace configuration

sclevine:


jncd:
Thanks! @sclevine

install.md
```diff
@@ -1423,3 +1423,74 @@ Use the following procedure to verify that the packages are installed.
...
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: registry-credentials
+  annotations:
+    secretgen.carvel.dev/image-pull-secret: ""
+type: kubernetes.io/dockerconfigjson
+data:
+  .dockerconfigjson: e30K
```
evankanderson, line 1452: These seem like they are redundant (they will be filled with the same credentials).

install.md
```diff
@@ -1423,3 +1423,74 @@ Use the following procedure to verify that the packages are installed.
...
+kind: Role
+metadata:
+  name: kapp-permissions
+  annotations:
+    kapp.k14s.io/change-group: "role"
+rules:
+  - apiGroups:
+      - serving.knative.dev
+    resources: ['services']
+    verbs: ['*']
```
evankanderson, line 1476: Worth a note that this might also need to cover resources like service bindings if in use?

evankanderson (commented):
Awesome, thanks for this!

I wrote up something similar in #183 -- do you think we should have both, or link to one from the other?

a-peek4:
@evankanderson @sclevine Please take a look at the comments and make a new PR if there are any doc changes to be made. Thank you!

PR #182, opened 2021-10-05
Extend timeout for CNR install to 30 minutes

evankanderson:
Per user reports, sometimes it takes more than 15 minutes to install the CNR package due to a large number of images and occaisonaly registry slowness / errors. This manifests as a reported timeout followed by subsequent installation success when verifying the install.


jncd:
Thanks @evankanderson

PR #199, opened 2021-10-05
Add individual docs for developer conventions and default supply chain

kiwi-bui:


jncd:
Thanks @kiwi-bui 

Do you have a tech writer on your team?

PR #204, opened 2021-10-05
Update install-intro.md

jessie-norden:
Changed instructions from passive voice to active voice


jncd:
Thanks @jessie-norden

PR #205, opened 2021-10-05
Add release notes for beta 2

sclevine:


jncd:
Thanks @sclevine 

I'm so glad that you're doing the release notes.

Please remember, no "please". Our word list says about "please": 
> Don't use. The exception is if you are asking the reader to do you a favor. For example, "Please report any security breaches to Pivotal Technical Support."

PR #206, opened 2021-10-05
Add clarification about configuring Java extension pack dependency

lisamburns:


jncd:
Thanks @lisamburns 

I tried a little _in situ_ editing. My top tips for today: 
Use simple verb tenses. You can usually use the present instead of the future if you mean soon. Only use the future tense if you can specify a time —  like after four hours you will be prompted. 
Simple verb tenses are so nice for ESL readers.

PR #163, opened 2021-10-04
Update metadata store step with creating namespace and secret and note to indent cert by two spaces

rxl7906:

PR #207, opened 2021-10-05
Add explicit note to say which certificate sample this instruction is referring to

rxl7906:
Follow up from @jncd 's comment https://github.com/pivotal/docs-tap/pull/163#issuecomment-933763083. This change is to make it more explicit to the user.


jncd:
Sorry @rxl7906 

I still don't really get the link between the `kubectl get` command and the spaces. I'm trying a different approach. See screenshot. Does this work for you? It seems that you need to worry about the spaces when you create the YAML, no?
Screen Shot 2021-10-05 at 2 44 15 PM

PR #208, opened 2021-10-05
Remove GKE from supported Kubernetes clusters

rnandi:


jncd:
Thanks @rnandi 

I'm so happy to see that line go. (The note only made sense if you already knew what it meant.)

PR #209, opened 2021-10-05
Add note about K8s version, remove abbreviations

sclevine:


jncd:
Thanks for spelling out all those abbreviations, @sclevine

PR #211, opened 2021-10-05
Clarify SupplyChain secret instructions

sclevine:


jncd:
Thanks @sclevine 
Yes, there's something really wrong with the gather the schema step.
I think when the dust settles, we should get rid of the General Procedure and just give specific instructions for each of the packages. (The general procedure kinda worked for beta-1 when there were only three packages but now the scrolling and the variations are too much.)

PR #214, opened 2021-10-06
add convention-server component

heyjcollins:
updating the nav to point to the page I'll be creating shortly at `convention-service/about.md`


jncd:
Thanks @heyjcollins

PR #215, opened 2021-10-06
Added GKE support back, updated kapp-controller version, Removed sect…

dhadukk:
Added GKE support back, updated kapp-controller version, Removed section set and verify Kubernetes cluster configurations, and combined it with prerequisites


jncd:
wow, so many last minute changes.
Thanks @dhadukk

PR #224, opened 2021-10-06
Additional RBAC Rules

nebhale:
Previously the collection of RBAC rules were missing a couple of resources that were required to make TAP work for users.  This change updates the RBAC instructions to add those rules and roles.


jncd:
@nebhale

Issue #79, opened 2021-09-30
Secret creation for API Portal is not required


msrinivasssa:
In [https://github.com/pivotal/docs-tap/blob/main/install.md#-install-api-portal](url), what is the need for creating the secret? if this is to pull the imgpkg bundle from tanzunet, it is already taken care by creating imagepullsecret in the earlier sections.
if this is created for some other reason, please specify where this secret will be used in later stages to provide some clarity to the user.

jncd:
Hi @fombico 
Can you resolve this issue with @msrinivasssa ?

@fombico wrote this portion of the docs.
@msrinivasssa is our SME (subject matter expert) with overall responsibility for this page.

fombico:
The secret for pulling the API portal image is called `api-portal-image-pull-secret`.

msrinivasssa:
HI @fombico Thanks for the clarification. what i don't understand is why we need a separate secret to pull the API portal images?
In the install docs, there is a section where we create a registry secret for tanzunet to pull images for the TAP components. This secret will be exported to all namespaces using secretgen-controller + kapp-controller. All the component teams are asked to create a placeholder secret for imagepull secret in the package CR so that the secret will be applied at run time.

Does API portal has that placeholder or no in the package CR?
if it already has the placeholder what is the need for this `api-portal-image-pull-secret`? is it safe to remove the step to create secret from the doc.
if the placeholder secret is not present in the package CR, please add it and remove the step to remove the step to create `api-portal-image-pull-secret`

dhadukk:
@msrinivasssa  Placeholder secret should not be part of package CR. Instead it should be part of API portal. I believe they already have place holder secret in API portal bundle.

msrinivasssa:
@dhadukk In that case, lets remove the additional secret creation step in the doc, It creates unnecessary confusion to the user.

fombico:
@msrinivasssa There is a placeholder secret in the package CR, named `api-portal-image-pull-secret`. This is the same secret the deployment is looking for. 

I'm uncertain as to how secretgen-controller will populate this, unless the user knew beforehand to create `api-portal-image-pull-secret` elsewhere and made it available with a SecretExport.

dhadukk:
[...]

PR #227, opened 2021-10-06
Remove step on API portal to remove image pull secret

fombico:
This step should have been removed in my previous PR https://github.com/pivotal/docs-tap/pull/188 to address #79.

However, when the commit titled _Revert "Update steps to install Application Live View"_ 
(sha `5d6a85d0b9fd332dd95d33ac57b721d673077b64`) went in,  it put this step back.


jncd:
Thanks @fombico 
Glad you noticed that. 
(A new engineer made a big oopsie commit without a PR so I had to revert in a hurry.)

PR #228, opened 2021-10-06
TANZUSC-455 Adds CNR namespace clarification

Petahhh:
* the namespace is not identified for this step in the CNR docs
* users should do this for every namespace they intend to deploy
knative services

Co-authored-by: David Solomon <solomond@vmware.com>


jncd:
Thanks @Petahhh

PR #231, opened 2021-10-06
Add prereq section for supply chain scan for metadatastore, fix cert comment, use proper present tense

rxl7906:
Summary:
* Add prereq section for supply chain scan to say metadata store is a preqreq.
* Remove duplicate cert comment
* Use less "-ing" in verbs in the running-scans.md because https://docs-staging.vmware.com/en/VMware-Tanzu-Application-Platform/0.2/tap-0-2/GUID-install.html this page seems to use present tense without "-ing"
* Fix headers in runnings-scans.md

### DIFF
--- a/install.md
+++ b/install.md
@@ -1179,6 +1179,9 @@ Ensure both are installed.
 
 To install Supply Chain Security Tools - Scan (Scan Controller):
 
+**Prerequisite**: Supply Chain Security Tools - Store installed on the cluster.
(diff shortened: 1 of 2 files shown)


rxl7906:
@jncd you're right, I need to change the link to the prod site. However the install-supply-chain-security-tools store does not exist yet in prod. Can I assume it will be this? https://docs.vmware.com/en/VMware-Tanzu-Application-Platform/0.2/tap-0-2/GUID-install.html#install-supply-chain-security-tools--store-15

rxl7906:
@jncd ignore my comment. I figured out how to properly set the link and made the change here https://github.com/pivotal/docs-tap/pull/234

PR #232, opened 2021-10-06
TANZUSC-456 Update CNR install example

Petahhh:
* In this step of CNR installation, using `default`
for the service account is the straight forward
way to complete the installation

Co-authored-by: David Solomon <solomond@vmware.com>


jncd:
Thanks @Petahhh

PR #269, opened 2021-10-15
Update install-general.md

msrinivasssa:


msrinivasssa:
> Hi @msrinivasssa
> 
> I'm merging this — looks like it needs some editing.
> 
> What does " (**BETA ONLY SUPPORT**) " mean for TKG v1.4?
> 
> I thought that everything was still in beta? (Or do you mean that there is a beta version of TKG v1.4 and only it works with TAP?)
> 
> Because TAP is all beta for now, I'll delete " (**BETA ONLY SUPPORT**) "

@jncd the intent is to alert the user "not to use this TKG cluster for any other production workloads" Since we're using a workaround to upgrade the kapp-controller, there is no guarantee that their workloads will function properly on this cluster. This cluster should be a throw away cluster. only for beta testing purposes.

PR #270, opened 2021-10-18
add help for prev installs in plugin section

heyjcollins:
folks are getting caught up in the install plugins step and not knowing what to do.
added a reference to error that occurs and pointer to follow the directions in the "update the Tanzu CLI" section to resolve the problem.


jncd:
Screen Shot 2021-10-18 at 4 19 43 PM

PR #271, opened 2021-10-19
Update install-tkg.md

msrinivasssa:


RichardJJG:
I've already applied this fix, thanks

PR #272, opened 2021-10-19
Update app accelerator install

trisberg:
- add source controller dependency
- move app accelerator after source controller
- add new config options


jncd:
Thanks @trisberg 

Just to confirm, I've merged this into the main branch which is for TAP v0.3. This change will not appear in the v0.2 docs.

trisberg:
yes, just TAP beta 3

PR #277, opened 2021-10-20
Updated Tanzu Application Platform GUI Naming

krisapplegate:
Changed Tanzu Application Platform UI to Tanzu Application Platform GUI


jncd:
Thanks @krisapplegate 

I wish we could get away with the TAP abbreviation, but we've already got caught once in beta-2 and had to spell them all out.

PR #281, opened 2021-10-21
Point to OSS Cartographer docs.

emmjohnson:
We do not want to maintain two sets of docs right now.


jncd:
@emmjohnson and @RichardJJG 

I'm merging this now; to give you a head start tomorrow. (I hope that's okay.)

PR #375, opened 2021-10-29
Add example Component + Workload to Workload Visibility docs

krisapplegate:


HenryBorys:
Hi @krisapplegate. Thanks! I've merged this. Just a heads up: I don't believe workload should be capitalized, except when it starts a sentence.  The writer editing the story will sort that out, but the added example is great. Thanks again.

PR #376, opened 2021-10-29
Removing references in internal teams

C10U:
removed references to internal teams in public documentation


HenryBorys:
Thanks @C10U

PR #378, opened 2021-10-29
Updating install docs

alexbarbato:
Reflecting all of the supply chains installed.


alexbarbato:
I need to validate with Mandy if this is even right

HenryBorys:
Thanks @alexbarbato. Merged, and should be live shortly.

PR #380, opened 2021-10-29
simplify "bounce the pod" instructions

jamieklassen:
Since our pods (for the time being) have a consistent `app=backstage` label you don't need an extra step to find which one to delete!


HenryBorys:
Thanks @jamieklassen . I merged your PR. Should be live soon.

PR #381, opened 2021-10-29
complete overhaul of Tanzu CLI install section

heyjcollins:
based on all the feedback from bug bash, this is an attempt to resolve the many issues associated with installing the Tanzu CLI.

## PLEASE CHECK 
- that the links pointing to TKG install and TCE install at the bottom of the file will resolve properly when on docs.vmware.com - can you update if needed?


HenryBorys:
@heyjcollins Thanks. I've merged your PR. Should be live soon.

PR #385, opened 2021-10-29
updating links to point directly to next step

heyjcollins:
once the TKG/TCE-specific instructions for the Tanzu CLI have been followed, the next step is to use the CLI to install TAP on TKG/TCE.
This change updates the pointers to deep-link directly to that part of the installing TAP on TCE/TKG pages.

## PLEASE
make sure these updated links will work properly on docs.vmware.com and if not, update accordingly -- thanks so much~!


HenryBorys:
@heyjcollins Thanks, Josh. We'll make sure the links work.Should be live soon.

PR #388, opened 2021-10-29
Update scst-store docs 

julia-pu:
Updated cli docs and api.md with new health check endpoint


HenryBorys:
@julia-pu Thanks, Julia. Merged. Should be live soon!

PR #389, opened 2021-10-30
Surround username and password with quotes

cschuyle:
This to remind the reader to put the username and password in quotation marks, since if a value contains certain characters, the error message is not very helpful:

Error: accepts 1 arg(s), received 2
Error: exit status 1


RichardJJG:
Thanks, @cschuyle!

PR #390, opened 2021-10-30
minor updates to inline links/header content

heyjcollins:
removed the reference to beta 2 for Updating the Tanzu CLI from previous TAP release so the instructions will remain durable through beta N and GA+
minor content updates as well


RichardJJG:
Thanks, @heyjcollins!

PR #391, opened 2021-10-30
fix inline links and update install instructions

heyjcollins:
updated instructions for convention service, source controller, and spring boot convention.


RichardJJG:
Thanks, @heyjcollins!

PR #395, opened 2021-10-31
Update getting-started.md

odedia:
Since we're documenting how to install the full profile in the installation docs, it makes more sense to add "install_tekton: true" to the tap-values.yaml and let it install automatically.


RichardJJG:
Thanks, @odedia!

PR #396, opened 2021-11-01
Update dependencies for default supply chain

Samze:
Add note that you will also need service bindings and services toolkit if using service references as part of a workload declaration.

Please reformat if needed.


RichardJJG:
Thanks, @Samze!

PR #397, opened 2021-11-01
Added TAP-GUI Auth note.

krisapplegate:


RichardJJG:
Thanks, @krisapplegate!

PR #402, opened 2021-11-01
Update remaining top level occurences of Services Control Plane Toolkit

edwardecook:
I wasn't sure about adding in the version for Services Toolkit in release-notes.md as it hasn't technically been added in although we are pretty certain that it will be v0.4.0.


RichardJJG:
Thanks, @edwardecook!

PR #404, opened 2021-11-01
Mention the LoadBalancer requirement in the prerequisities

evankanderson:
Several users hit problems where they were attempting to install on a cluster with no LoadBalancer type, typically on VSphere. Mention this requirement, using the same language as here: https://docs.vmware.com/en/Cloud-Native-Runtimes-for-VMware-Tanzu/1.0/tanzu-cloud-native-runtimes-1-0/GUID-install.html#prerequisites-0


RichardJJG:
Thanks, @evankanderson!

PR #407, opened 2021-11-01
Add default-supply-chains.md to toc.

emmjohnson:


jncd:
looks good
Thanks so much @emmjohnson

PR #408, opened 2021-11-01
Update install-general.md

andreizimin:
Added internet browser as a prerequisite. Called out no Safari support for TAP GUI


russeta:
@andreizimin Thanks. Carol

PR #412, opened 2021-11-02
TAP GUI Addition to Profile Install

krisapplegate:
We've added some additional documentation to the TAP profile install in order to make it easier when using profiles. We've also now updated the Getting started to account for the inclusion of the Catalog file in the accelerator bundle.


RichardJJG:
Thanks, @krisapplegate!

PR #413, opened 2021-11-02
Added a call to action at the end of TAP GUI

krisapplegate:


RichardJJG:
Thanks, @krisapplegate!

PR #425, opened 2021-11-03
Update link to new Services Toolkit docs url

edwardecook:
Note that currently this link is not active but it will be once it has been bumped from staging.


RichardJJG:
Thanks, @edwardecook!

PR #435, opened 2021-11-04
ootb: point at build.5; add notes about one supply chain in the cluster at a time

cirocosta:
hi folks,

with the updates brought in via 0.3.0-build.5 (see https://gitlab.eng.vmware.com/tap/tap-packages/-/merge_requests/146 - not merged yet), we had to update a couple things in the guides over here.

highlight: unfortunately, with the focus on the _developer_ experience of not having to target a specific supply chain (e.g., not having to specify `--type=web` vs `--type=web-test` but instead having cartographer "figuring out" which one to use under the hood, something to be implemented), we had to rollback support for having all 3 supply chains coming out of the box in the profiles and going back to documenting how operators should uninstall previous supply chains so the matching always works.

ps.: this should only be merged after build.5 is available (given the changes in `--type=` for `tanzu apps workload create` commands).

thank you!

/cc @jwntrs

### DIFF
--- a/getting-started.md
+++ b/getting-started.md
@@ -28,7 +28,9 @@ Before getting started, ensure the following prerequisites are in place:
 
 2. Default kubeconfig context is set to the target Kubernetes cluster
 
-3. A developer namespace has been setup to accommodate the developer's Workload.
+3. The Out of The Box Supply Chain Basic is installed: see [Install default Supply Chain](install-components.md#install-ootb-supply-chain-basic).
+
+4. A developer namespace has been setup to accommodate the developer's Workload.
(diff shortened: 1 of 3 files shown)


cirocosta:
(force-push to resolve conflicts)

RichardJJG:
@cirocosta, the below is now missing from the docs:

```bash
    tanzu package installed delete ootb-supply-chain-basic
    tanzu package installed delete ootb-supply-chain-testing
```

Is that by design? I know this PR had merge conflicts, but whole commands tend not to get wiped out on purpose when resolving clashes.

PR #437, opened 2021-11-04
Add details for troubleshooting tbs

tylerphelan:
- Add section for when there are no build logs or no builds

### DIFF
--- a/troubleshooting.md
+++ b/troubleshooting.md
@@ -70,3 +70,30 @@ A common cause of this error is that the `tanzu package install` command is bein
 To fix this problem:
 
 1. Use `tanzu package installed update` command after the first use of `tanzu package install` command if you want to update the package.
+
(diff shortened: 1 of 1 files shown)


HenryBorys:
@tylerphelan  Hi Tyler. Re: the angle brackets in this line: ```bash
tanzu apps workload tail <workload-name> --since 10m --timestamp
```
@jncd  had told me we don't use those, even in code. However, she also noted an exception: "If the YAML file is one that customers have downloaded from TanzuNet, then they will see exactly those angle brackets..." so they're appropriate then. With that in mind, should they be used in this case? Thanks!

tylerphelan:
@HenryBorys I just wanted to represent the "name of the workload" as that argument

HenryBorys:
@tylerphelan Yep, I figured. Let me just confirm with one of the FTE editors whether they should stay in this case.

PR #441, opened 2021-11-05
Refactor CNR installation steps

Petahhh:
* removed copy and paste crust
* grouped optional steps under a single step


jncd:
Thanks @Petahhh 
I'm merging and will push this to the staging asap so you can take a look.

PR #442, opened 2021-11-05
Streamline Services Toolkit Advanced Use Cases

teddyking:
Hello! This PR replaces the "Component Overview" section of the Services Toolkit Advanced Use Case with a link to our own component docs.

Note: I have used a link to a URL that doesn't currently exist (https://docs.vmware.com/en/Services-Toolkit/0.4/services-toolkit-0-4/GUID-overview.html), but that will exist once the linked documentation goes live. Is this the preferred approach when linking to docs in PRs? Or would you prefer if I were to link to the docs on the staging site?

Thanks!
Ed

### DIFF
--- a/getting-started.md
+++ b/getting-started.md
@@ -1010,14 +1010,12 @@ One of the out of the box supply chains we are working on for a future release w
 
 * [Observing and Troubleshooting](scst-scan/observing.md)
 
-
-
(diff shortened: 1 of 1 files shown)


jncd:
THanks @teddyking 

I'm merging and will push this to the staging asap so you can take a look.

I think it's best to do as you have and cross-reference the site that will be live on Monday.

PR #456, opened 2021-11-05
Clean up pre-req

mgibson1121:


BarrAndrea (changes requested):
Please change the passive tone of "If available" to "When available" on line 25.

BarrAndrea:
@mgibson1121  I am merging this and making that phrasing change.

PR #469, opened 2021-11-08
Edited the service-type instructions

krisapplegate:


RichardJJG:
Thanks, @krisapplegate

PR #485, opened 2021-11-10
Remove confusing example for password value

tylerphelan:
- I thought this would be a good way to represent how to fill
  out this value but a user was confused by it so I am removing the example.


RichardJJG:
Thanks, @tylerphelan!

PR #486, opened 2021-11-10
Move default namespace up

waldirmontoya25:
Move the recommendation to use the default namespace up, before the actual procedure.


RichardJJG:
Thanks, @waldirmontoya25!

PR #491, opened 2021-11-12
Convention controller reference and creation

odinnordico:
Added documentation of how to create conventions, this included the needed models.


RichardJJG:
Thanks, @odinnordico!

PR #494, opened 2021-11-12
Changes to parameters and also requirements of scan component

bsoroushian:
Signed-off-by: Behrouz Soroushian <bsoroushian@vmware.com>


a-peek4:
Thanks @bsoroushian Does this apply to 0.3 and 0.4?

jncd:
^^ @bsoroushian We're assuming that this just goes in the 0.4. Let us know if that's not so. Thanks.

PR #496, opened 2021-11-15
Add default read-write user flag to docs-tap

Pivotal-Jeff-Jun:


a-peek4:
Thanks @Pivotal-Jeff-Jun Does this change also apply to 0.3?

Pivotal-Jeff-Jun:
@a-peek4 This doc change is for 0.4.

PR #499, opened 2021-11-17
Removed duplicate bullet point

ciberkleid:
Removed duplicate bullet point recommending the use of a paid registry to avoid rate-limiting.


RobC55:
@ciberkleid Thanks!
Does this apply to 0.3 docs as well as 0.4 docs?

PR #503, opened 2021-11-17
Added docs for spring-boot-conventions including troubleshooting instructions

suarezjulian:


RobC55:
@suarezjulian Thanks.
Will pass along to @HenryBorys and @BarrAndrea (tech writers for "Conventions").

PR #507, opened 2021-11-18
application.yaml -> accelerator.yaml

rhardt-pivotal:
I'm pretty sure we mean 'Using accelerator.yaml'


RichardJJG:
Thanks, @rhardt-pivotal!

PR #510, opened 2021-11-18
Adding troubleshooting section for Source Controller

warango4:


warango4:
@russeta The PM is @heyjcollins. Can you point where it needs editing? I'm working for Framework Team under Developer Productivity team, we're responsible for Apps Plugin, Service Bindings, Convention Controller and Source Controller, and since I saw the other components listed together, I added this one under the last that was commmited. But if you have any suggestion, I'll happily change according to what you're expecting. Thanks :)

jncd:
HI @warango4  Thanks. 
I'm handing this PR over to @BarrAndrea and @HenryBorys 
They are writers who are already getting to know the Developer Productivity space. Probably they'll reach out to you for a chat.

PR #514, opened 2021-11-22
Update to approved agg role label and finalizer key

gmrodgers:
For the beta 4 release.  Any place with  services.apps.tanzu.vmware.com/aggregate-to-resource-claims: "true" should be replaced with resourceclaims.services.apps.tanzu.vmware.com/controller: "true"


jncd:
Thanks, @gmrodgers 
Did you do a global search for `services.apps.tanzu.vmware.com/aggregate-to-resource-claims`? Or might there be others lurking?

PR #516, opened 2021-11-22
Add update to resource claim cluster role as is now required

edwardecook:
This is required as part of the exclusive resource claims feature.

This is a feature for tap-beta4, it doesn't need to be ported back to previous versions.


jncd:
Thanks @edwardecook 
I really appreciate that you call out the beta-4 only features.

PR #518, opened 2021-11-22
Remove comment that shows online

odinnordico:
Removing comment that shows online for spring boot conventions


jncd:
Thanks @odinnordico

PR #520, opened 2021-11-22
Add video demoing SCST - Store

Manifaust:
I upload a demo I made to YouTube and I embedded it onto the overview.


jncd:
Cool! @Manifaust 

Just remember that you have to treat videos like doc — so you're responsible for reviewing it periodically and making sure that all the info is still correct and that the video doesn't have to be updated when a new release of scst-store gets released.

jncd:
Hi Tony, @Manifaust 
The video is now embedded: https://docs-staging.vmware.com/en/Tanzu-Application-Platform/0.4/tap/GUID-scst-store-overview.html

But it's just there — still some chores that need to be done. 
1) Introduce the video, describe what's in it so as a reader I know whether I want to grab my headset and watch this video or skip it and keep reading.
2) Need some alt-text so that a screen-reader knows what to say when it encounters the video.

PR #529, opened 2021-11-24
Add blurb on read-only cluster role for Metadata Store

xtreme-debbie-chen:


RichardJJG:
Thanks, @xtreme-debbie-chen!

PR #536, opened 2021-11-29
Improve instructions for using Dockerhub with supply chain

Manifaust:
The previous instructions were unclear or incorrect.


RichardJJG:
Thanks, @Manifaust!

PR #539, opened 2021-11-29
fix broken link to default supply chain on TanzuNet

heyjcollins:


jncd:
Thanks @heyjcollins 
I'll make the same fix to the v0.3 docs.

PR #542, opened 2021-11-30
Updated the tap-values.yml for contour ingress

msrinivasssa:
Updated the `tap-values.yml` with the values for contour.


Lucas-Hanson:
Thanks

PR #555, opened 2021-12-02
update spring boot convention name

heyjcollins:
convention descriptions is more helpful than "resources" so I updated the name.
reversed order of the sub-items as well.


RichardJJG:
Thanks, @heyjcollins!

PR #561, opened 2021-12-02
Add ceip key to yaml sample - call out install fails if not set to true 

mgibson1121:


mgibson1121:
Final form of this doc will have a link to the CEIP page.  Thinking about adding a forwarding URL in the meantime.

PR #562, opened 2021-12-02
Small tweaks to language

mgibson1121:


jncd:
Nice! I fixed the passive voice; Richard would have caught it.

PR #566, opened 2021-12-02
ootb: update installation and detailed sections

cirocosta:
hi folks!

sorry for the huge PR - the majority of the changes are related to ootb-*
specific additions under `./scc` and updates to individual component 
installation under `install-components` that were either lacking or that
needed some updates.

- install-components: installation instructions for some ootb-* products
were missing and other needed to be updated with the new requirements of
having to have `delivery-basic` and `templates` as prerequisites for
all, and `testing` not being compatible with `testing-scanning` at the
same time.

- scc: add detailed entries about each package in each file - we haven't
been updating those files very often, but, now we'll be keeping them in
sync with internal docs on every release

- getting-started:  in the new supplychains included in ootb 0.4.0-build.2, 
pipelines match based on label selection (thus, the addition of the label 
in the tekton/Pipeline), and supplychains that make allow running tests
match on workloads that also include the `has-tests` label.

with these changes, I believe we can improve the `getting-started` to
point folks at the `scc/ootb-*` pages where we can get more in-depth
about what's possible / details of each supplychain, but i think it'd be
better to do that in a follow-up pr.

please let me know if there's anything I can help with - I understand it can
be quite annoying to go through large PRs like this, but happy to help.

thank you!


a-peek4:
Thanks @cirocosta

jncd:
Hi @cirocosta 
Deleting scc/default-supply-chains.md broke the build, I think. So I removed it and added all the files in the scc directory to the toc in alphabetic order. You should take a look and reorder as appropriate. 
https://github.com/pivotal/docs-tap/commit/9411eb21f1f85c9ff3672d2d3b6c8c90357a0557

Note: We've recently switched from title case to sentence case for our headings: i.e. now "Out-of-the-box delivery basic" instead "Out-of-the-Box Delivery Basic" 

(Although now that I write it, I don't think that title makes sense.  🤣 Should start with an 'ing' word to describe what you do with delivery basic. "Delivering basic out-of-the-box _something_"? )

PR #571, opened 2021-12-03
Getting Started Changes

krisapplegate:


jncd:
Thanks @krisapplegate 
I'll merge and let you know when it's on staging. 
(My first guess is that we'll want to indent those images by four spaces so that the numbering doesn't reset. We'll see.)

jncd:
@krisapplegate It's on staging. Unfortunately, I was right about the indents.

PR #589, opened 2021-12-03
fix versions, install steps, add services plugin

heyjcollins:
* corrected the versions of the plugins that should be installed (in the output from the various `tanzu plugins list` verifications)
* added the installation of the new `services` plugin
* updated instructions to properly disable the `context-aware-cli-for-plugins` feature AFTER the CLI core is updated and BEFORE plugins are installed, IF the CLI core was upgraded from a version prior to `v0.10.0`
* made some minor copy edits

### DIFF
--- a/install-general.md
+++ b/install-general.md
@@ -362,20 +362,21 @@ To perform a clean installation of the Tanzu CLI plugins:
     Expect to see the following:
 
     ```
-     tanzu plugin list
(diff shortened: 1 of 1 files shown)


jncd:
Screen Shot 2021-12-06 at 12 27 12 PM

^^ @heyjcollins

PR #594, opened 2021-12-03
update tap profiles install doc

katmutua:
add a default `ceip_policy_disclosed` and clarification on type expected


a-peek4:
Thanks! @katmutua

PR #597, opened 2021-12-03
Additional installation step added to intro page

amakhija:
Added tanzu dev tools installation step as step 7 on install intro page


a-peek4:
Thanks @amakhija

PR #598, opened 2021-12-03
formatting for console copying

krisapplegate:


jncd:
@krisapplegate 
I'm sure this didn't render as you hoped. I made an attempt to change it. Perhaps you should check it.

https://docs-staging.vmware.com/en/Tanzu-Application-Platform/0.4/tap/GUID-install.html

Screen Shot 2021-12-03 at 7 06 41 PM

a-peek4:
@krisapplegate I made some changes to this, LMK if it looks good to you. Thank you!

PR #604, opened 2021-12-06
Fix typo to be proper variable assignment

rxl7906:
This looks like a typo. This needs to be a proper variable assignment.


jncd:
It sure does look like a typo. Thanks @rxl7906

PR #605, opened 2021-12-06
make file delete more exact for CLI removal

heyjcollins:
folks may be storing other TAP-experimentation-related files in $HOME/tanzu --- so the current delete would remove all of those (potentially making people sad/upset). Updating so it deleted $HOME/tanzu/cli and below should help address that.


RichardJJG:
Thanks, @heyjcollins, sounds good!

PR #608, opened 2021-12-06
Explain how to create image patterns and wildcard rules in our image policy

elfotografo007:
Signed-off-by: Andrés Torres <andrest@vmware.com>

### DIFF
--- a/scst-sign/configuring.md
+++ b/scst-sign/configuring.md
@@ -217,6 +217,29 @@ namespace and add the secret names from step 1 to the `imagePullSecrets` section
 
     Add additional secrets to `imagePullSecrets` as required.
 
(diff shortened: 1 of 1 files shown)


jncd:
Thanks @elfotografo007 

I'll merge and take a look. 
BTW, we've recently changed the way we write headings to match the new VMware style. So "Providing Image Name Patterns" => "Providing image name patterns". Except that we don't start headings with gerunds and we don't use verbs in sections unless the section contains a numbered procedure. [🤣 , so many rules, just like coding]
I'll use "Image name patterns" for the heading. 😃

elfotografo007:
Thanks for the information.

PR #613, opened 2021-12-07
Be clear that insight CLI is beta

Manifaust:
And there isn't necessary going to be a CLI plugin to replace it.


jncd:
Thanks @Manifaust 

BTW, I standardized on "Insight CLI" rather than `insight` CLI. There was a mix in the doc.

PR #614, opened 2021-12-07
[beta-3] Clarify that insight CLI is in beta

Manifaust:
Remove reference that we're making a CLI plugin


jncd:
Thanks @Manifaust 
I pushed to production: https://docs.vmware.com/en/VMware-Tanzu-Application-Platform/0.3/tap-0-3/GUID-scst-store-overview.html#recommended-3

PR #619, opened 2021-12-07
Updating fluxcd source-controller documentation

EstebanFS:
- Updating documentation so it can use the TAP fluxcd.source.controller.tanzu.vmware.com package instead of the manual installation

### DIFF
--- a/install-components.md
+++ b/install-components.md
@@ -67,26 +67,78 @@ cert_manager and FluxCD Source Controller are installed as part of all profiles.
 * **FluxCD source-controller**:
     Install FluxCD source-controller using the following procedure.
 
-     1. Create the namespace `flux-system`.
+    1. List version information for the package by running:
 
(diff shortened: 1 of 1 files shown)


jncd:
Hi @EstebanFS 
Thanks for this PR. It basically looks very nice except we are worried that you deleted the instruction to create the namespace:

"Create the namespace `flux-system`."
```
kubectl create namespace flux-system
```

Yet, this namespace is required later in the procedure. 
     ```
        kubectl get pods -n flux-system
        ```

Maybe you want to edit this PR a little more?

EstebanFS:
Additional to the regular fluxcd configuration, we are adding this two descriptions, that replaced the creation of the namespace and the RoleBinding commands on the documentation.

image

Hope that is ok, let me know if there is anything we could add to make it clearer

PR #630, opened 2021-12-08
Tap GUI page for Application Accelerator

aclement:


tap-gui/plugins/application-accelerator.md
```diff
@@ -0,0 +1,75 @@
...
+## Creating the project
+
+You will now see the system generate the project. The Task Activity page will show the steps being taken together with detailed logs on the right side.
+
+![Task activity during project creation](./images/aa5_taskActivity.png)
+
+Once complete you can `Explore Zip File` which is similar to the `Explore` option we could have done earlier, or you
+can click `Download Zip File` to grab the project and download it as a .zip.
+
+## Developer your code
```
trisberg, line 70: Should this be "Develop"?

trisberg:
LGTM except for that one comment above

tap-gui/plugins/application-accelerator.md
aclement: bah, you caught me working too quickly ;) nice catch.

PR #643, opened 2021-12-10
Doc updates for VSEE-349

chaomonica:
VSEE-349: More Docs Edits
- Update Grype release version
- Update CRD names

### DIFF
--- a/scst-scan/explanation.md
+++ b/scst-scan/explanation.md
@@ -5,16 +5,16 @@ from [Installing Individual Packages](../install-components.html#install-scst-sc
 the following Custom Resource Definitions (CRDs) are now available:
 
 ```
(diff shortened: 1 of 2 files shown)


YugangW:
Hi @chaomonica, Thanks for the PR. I've merged it to the main branch, which will be applied to TAP 1.0. Please let me know if you believe this PR should also be applied to TAP 0.4 as well. Thanks!

RichardJJG:
@YugangW, these changes are specifically for docs about SCST - Scan, which is a component separate from TAP itself. SCST - Scan has its own [release notes](https://docs.vmware.com/en/VMware-Tanzu-Application-Platform/0.4/tap/GUID-scst-scan-release-notes.html). 
Therefore, we don't need to check if the changes "apply" to any particular version of TAP because these docs aren't really about TAP itself.

PR #644, opened 2021-12-10
remove `sudo` from CLI install on mac

heyjcollins:
sudo isn't required on mac so removing.

### DIFF
--- a/install-general.md
+++ b/install-general.md
@@ -263,7 +263,7 @@ To install the Tanzu CLI on a Mac operating system:
 6.  Install the CLI core:
     ```
     cd $HOME/tanzu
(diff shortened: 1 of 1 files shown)


YugangW:
Hi @heyjcollins, Thanks for the PR. I've merged it to the main branch, which will be applied to TAP 1.0. Please let me know if you believe this PR should also be applied to TAP 0.4 as well. Thanks!

heyjcollins:
This change can be applied to TAP beta3+
________________________________

YugangW:
Thank you @heyjcollins. This change has been applied to beta 4 as well.

PR #658, opened 2021-12-15
Added known issue for java source scan.

nedenwalker:
This is relevant to all versions of scanlink.


RichardJJG:
Thanks, @nedenwalker!

PR #662, opened 2021-12-15
Update scan results with additional fields for VSEE-341

chaomonica:


YugangW:
Hi @chaomonica, Does this change apply to any earlier versions of SCST-Scan?

chaomonica:
@YugangW No, they do not need to be added into earlier versions of SCST-Scan. Thanks!

PR #672, opened 2021-12-17
Clarified known issues.

nedenwalker:
Just latest (ie 0.5.0 and 1.0.0)


RichardJJG:
Thanks, @nedenwalker!

PR #675, opened 2021-12-17
ootb: update components documentation

cirocosta:
hi folks!

here I update install-components and `scc/`-specific documentation to account for the changes that we brought in with beta5.

the biggest changes are:

- **install-components**: new params (`gitops.*`) for the `ootb-supply-chain-*` packages
- **install-components**: add ootb-delivery-basic (it was missing before)
-**install-components**: bump cartographer and all ootb- versions
- **scc**: up-to-date documentation on how to make use of the gitops params that an operator sets / gets overwritten by a developer

### DIFF
--- a/install-components.md
+++ b/install-components.md
@@ -17,10 +17,11 @@ For information, see [Installing part I: Prerequisites, EULA, and CLI](install-g
 + [Install Application Accelerator](#install-app-accelerator)
 + [Install Tanzu Build Service](#install-tbs)
 + [Install Supply Chain Choreographer](#install-scc)
++ [Install Out of the Delivery Basic](#install-ootb- delivery-basic)
 + [Install Out of the Box Templates](#install-ootb-templates)
 + [Install Out of The Box Supply Chain Basic](#install-ootb-supply-chain-basic)
 + [Install Out of The Box Supply Chain with Testing](#install-ootb-supply-chain-testing)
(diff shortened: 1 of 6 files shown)


jncd:
Wow, loadsa changes @cirocosta 
I'll merge and take a look. 
You can see your changes in this build of the docs too: https://docs-staging.vmware.com/en/Tanzu-Application-Platform/1.0/tap/GUID-overview.html

jncd:
Hi @cirocosta It's the first time I've looked at these pages. Someday it will get a complete edit, I hope. (just not by me, haha.)  I tried to make some small fixes as I read through.  Going forward, you could use a text editor with a spelling and grammar checker, that would help. Also, you could preview the text in a markdown viewer before submitting.

I did get stuck on the mix of "we"s and "you"s. I suspect that really it should just be "you" consistently and I tried to change them. For example, from 
"we'd see kubernetes configuration being pushed to `git@github.com/foo/tanzu-java-web-app.git`"
to "You see the Kubernetes configuration pushed to `git@github.com/foo/tanzu-java-web-app.git`.

There are some sections that I didn't touch because I couldn't understand them. For example, 
"Differently from local iteration, with the GitOps approach we end up at the end
of the supply chain having the configuration that got created by it pushed to a
git repository where that's persisted and used at the basis for further
deployments."  It doesn't look like you wrote the original, but maybe you could help with a rewrite sometime in the future.

But, thanks for all your good work keeping these docs up to date.

PR #677, opened 2021-12-17
Section 1 related to TAP GUI

andreizimin:
Updated Section 1 - parts related to TAP GUI. Together with Sonja Ely.

### DIFF
--- a/getting-started.md
+++ b/getting-started.md
@@ -14,11 +14,6 @@ Welcome to the Tanzu Application Platform. This document guides you through gett
 
 To take full advantage of this document, ensure you have followed [Installing Tanzu Application Platform](install-intro.md).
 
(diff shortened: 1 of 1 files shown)


jncd:
@andreizimin  Don't tell anyone, but I merged this right away so you could take a look: 
https://docs-staging.vmware.com/en/Tanzu-Application-Platform/1.0/tap/GUID-getting-started.html

jncd:
Top tips: 
+ Keep line lengths short, around 110 chars. 
+ Start each sentence on its own line. Markdown only creates new paragraphs after a blank line.
+ We've changed our heading style to match the more modern style: sentence case not title case. So going forward, we only need to capitalize the first word in a title.

PR #679, opened 2021-12-17
Release notes for Store 1.0.1

Manifaust:
Also add comment about CLI compatibility.

This should not be merged with any other branch.

### DIFF
--- a/scst-store/cli_installation.md
+++ b/scst-store/cli_installation.md
@@ -1,6 +1,6 @@
 # CLI installation
 
-> **Note:** The Insight CLI is in beta and is separate from the Tanzu CLI. It still works with the final 1.0.0 version of Supply Chain Security Tools - Store.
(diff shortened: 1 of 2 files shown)


RichardJJG:
Could we have more detail for the below, please?
>Changed package name to `metadata-store.apps.tanzu.vmware.com`

I'm concerned the reader won't know from what specifically **will** break, what **could** break, how to prevent failures, etc, etc.

Manifaust:
@RichardJJG upon reflecting on your question, I think we might just want to remove that mention of a breaking change.

This breaking change has to do with breaking upgrades between beta 4 and GA. If the reader is a first-time user, they would not have any breaking changes. The note is really for users of beta 4.

RichardJJG:
@Manifaust, the warning is still necessary then, isn't it? A "breaking change" by definition applies to users of the previous version who might "break" something when they upgrade.

I get that most users of v1.0 will be first-time users, but we still want to assist the beta users who want to upgrade.

PR #684, opened 2021-12-20
Changes to Section III content for supply changes

sonjawriter:


a-peek4:
@sonjawriter I see all these changes on main already (probably because I just merged some of your other PRs) so I'm going to close without merging. Please let me know if I missed something!

PR #688, opened 2021-12-20
Changes to Application Accelerator

sonjawriter:


a-peek4:
@sonjawriter In general, I'd advise against using H4s (####) instead of H3s (###) because the H4s render in gray and we've gotten feedback that it's hard to read

PR #694, opened 2021-12-20
Fixing TOC for supply chain tools - store

miketeve:


a-peek4:
Hi @miketeve This change moves all the files under the Release Notes section - was that your intention? I don't think having to click Release Notes in the TOC to expand all these files makes sense

miketeve:
@a-peek4 yeah i agree that doesn't make sense but currently, our stuff wasn't nesting correctly.  i was actually copying what i was seeing with other teams in hope that this would work.

PR #703, opened 2021-12-21
Delete self-signed-registry-workaround.md

heyjcollins:
This workaround, in practice, doesn't work as expected and must be removed.


RichardJJG:
Thanks, @heyjcollins!

PR #706, opened 2021-12-21
Changed button format

andreizimin:
Which other branches should this be merged with (if any)? NA

Changed certain button format from **BUTTON** to `BUTTON`


RichardJJG:
Thanks for the effort, @andreizimin, but it's a deliberate style choice to embolden UI elements instead of using code tags.

PR #708, opened 2021-12-21
Document path to have two or more apps consume from same service instance.

gmrodgers:
- also add warning to cross-namespace path as there currently is no path to having two apps consume from the same service instance across namespaces.

Which other branches should this be merged with (if any)? Beta 4 and Beta 5

### DIFF
--- a/getting-started.md
+++ b/getting-started.md
@@ -1364,7 +1364,7 @@ users.
     rules:
     - apiGroups: ["bindings.labs.vmware.com"]
       resources: ["provisionedservices"]
-      verbs: ["get", "list", "watch"]
+      verbs: ["get", "list", "watch", "update"]
     ---
     apiVersion: rbac.authorization.k8s.io/v1
     kind: ClusterRoleBinding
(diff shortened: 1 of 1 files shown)


getting-started.md
```diff
@@ -1344,8 +1346,119 @@ RabbitMQ instance:
 1. Visit the URL and confirm the app is working by refreshing the page and checking
 the new message IDs.
 
+#### <a id='binding-multiple-apps-to-the-same-service-resource'></a> Binding multiple apps to the same service resource
+
+Currently, if there are two workloads both claiming the same service resource, then Resource Claims would prevent the second workload from doing claiming the resource.  To get around this, a `ProvisionedService` can be used.
```
jhvhs, line 1351: from ~doing~ claiming the resource.

getting-started.md
```diff
@@ -1344,8 +1346,119 @@ RabbitMQ instance:
...
+
+1. In `provisionedservice-reader.yaml`, ensure you have RBAC enabled for all users.
+The following example grants `get`, `list` and `watch` to all `provisionedservice` resources for all authenticated
```
jhvhs, line 1354: `watch` ~to~ on all `provisionedservice` resources ~for~ to all authenticated

jhvhs (changes requested):

RichardJJG:
Thanks, @gmrodgers!

PR #738, opened 2021-12-23
Improve 'Before you begin' section in Runtime Resource Visibility

obabativavmware:
- beta-4


jncd:
This PR arrived after the docs-freeze deadline of COB Dec 22nd.

HenryBorys:
@obabativavmware in this article, there are links to /docs-tap/vscode-extension/usage-getting-started.md, but I can't find that article online. Do you know anything about that page? It looks somewhat similar to https://docs-staging.vmware.com/en/VMware-Tanzu-Application-Platform/0.4/tap/GUID-vscode-extension-usage.html, but they're not really the same.

PR #739, opened 2022-01-02
Removed Beta 5 Workaround for GA

krisapplegate:


BarrAndrea:
@krisapplegate Thank you for this update. I am approving and merging this change.

PR #740, opened 2022-01-03
typo, will prevent successful supply chain deployment

cpage-pivotal:


BarrAndrea (approved):
I am approving and merging this pr.

PR #750, opened 2022-01-05
rm redundant convention service known issues entry

heyjcollins:
this change should be applied to TAP docs v0.5.x+

### DIFF
--- a/release-notes.md
+++ b/release-notes.md
@@ -17,10 +17,9 @@ This release has the following new features:
 
 This release has the following issues:
 
-- **Component: Convention Service** Convention Service does not currently support self-signed certificates for integrating with a private registry. Support for self-signed certificates is planned for an upcoming release.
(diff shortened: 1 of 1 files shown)


release-notes.md
```diff
@@ -17,10 +17,9 @@ This release has the following new features:
 
 This release has the following issues:
 
-- **Component: Convention Service** Convention Service does not currently support self-signed certificates for integrating with a private registry. Support for self-signed certificates is planned for an upcoming release.
 - **Application Accelerator:** Build scripts provided as part of an accelerator do not have the execute bit set when a new project is generated from the accelerator. To resolve this issue, explicitly set the execute bit by using the "chmod" command: `chmod +x <build-script>`. For example, for a project generated from the "Spring PetClinic" accelerator, run: `chmod +x ./mvnw`.
 
-- **Convention Service:** Convention Service does not currently support self-signed certificates for integrating with a private registry. Support for self-signed certificates is planned for an upcoming release.
+- **Component: Convention Service:** Convention Service does not currently support self-signed certificates for integrating with a private registry. Support for self-signed certificates is planned for an upcoming release.
```
scothis, line 22: we don't currently support custom CAs. "self-signed certs" is a subset of custom CAs

```suggestion
- **Component: Convention Service:** Convention Service does not currently support custom certificates for integrating with a private registry. Support for custom certificates is planned for an upcoming release.
```

PR #751, opened 2022-01-05
Drop ProvisionedService resource

scothis:
We are not supporting the ProvisionedService resource as part of the TAP API. It will be removed in a future release.


YugangW:
Thanks for the PR, @scothis. Is this change applicable to the beta-4 branch as well?

scothis:
@YugangW it could be applied to beta4 as well, but that's not critical

PR #752, opened 2022-01-05
Drop binding multiple workloads to a service

scothis:
These workarounds are hacky and will break in future TAP releases. We would prefer to not need to support hacks moving forward that sidestep current platform limitations.


a-peek4:
@sonjawriter FYI

PR #754, opened 2022-01-05
Update sct sign tap beta version reference

katmutua:


a-peek4:
@katmutua Hi - can you verify that this should say "beta-4" even though this is for the GA version? I want to double check before I merge it

katmutua:
@a-peek4 Yes we are at this version `beta-4`.

PR #765, opened 2022-01-06
Reversed changes in #757 and added clarification

andreizimin:


RichardJJG:
Thanks, @andreizimin!

RichardJJG:
@andreizimin, does this apply to beta-4 also?

andreizimin:
Yes, it applies to Beta-4 also. Thanks.

PR #766, opened 2022-01-06
Install procedure for cert-manager and contour packages from TAP repo using PackageInstall CR

msrinivasssa:
Install procedure for cert-manager and contour packages from TAP repo using PackageInstall CR


krisapplegate:
As of last week, it looks like we're not going to be including any docs on TKG (since it's not yet supported). I'd remove the -TKG.md file and we should be good to go.

krisapplegate (changes requested):
Since we're now removing all TKG references, you can skip the TKG.md and just do the install-components.md changes.

PR #772, opened 2022-01-06
Add release note on SBOM issue

mingxiao:


RichardJJG:
Thanks, @mingxiao!

PR #778, opened 2022-01-07
 Adjusted resource requirements as well as Developer namespace refere…

krisapplegate:
…nces


YugangW:
Thanks for the PR, @krisapplegate. Do you want to apply these changes to beta-4 branch as well? Thanks!

PR #781, opened 2022-01-07
fixed spacing on indents for sample values files

krisapplegate:


animatedmax:
Thank you, @krisapplegate

PR #787, opened 2022-01-10
Services usecase table update

srikanthmanvi:


mpriamo:
This needs to get into 1.0

PR #795, opened 2022-01-11
Tap catalog/54 - Update OOTB version numbers to 0.5.1

taylorsilva:

Not sure which branch but sticking with the default `main`. ~@jwntrs any insight for which branch?~

```
Supply Chain Choreographer -> 0.1.0
Catalog -> 0.5.1
TAP -> 1.0.0
```

### DIFF
--- a/getting-started.md
+++ b/getting-started.md
@@ -526,7 +526,7 @@ To apply this install method, complete the following steps:
 3. Update the installed profile by running:
 
     ```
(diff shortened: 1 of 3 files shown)


a-peek4:
Thanks @taylorsilva

a-peek4:
@taylorsilva this is in 'main' branch for 1.1 right now, am I correct to assume it should also be cherry-picked to the 1.0 branch?

PR #797, opened 2022-01-12
2nd sentence had mixed up words

heyjcollins:
which really changed the meaning.


RichardJJG:
Thanks!

PR #809, opened 2022-01-17
Update the services toolkit documentation link

jhvhs:
Update TAP 1.0 documentation to include 0.5.0 of services-toolkit, as this is the version being shipped.

also needed in main.


RichardJJG:
Thanks!

PR #821, opened 2022-01-21
Update known issues for 1.0.1

lisamburns:
- Note looks like none of the known issues from 1.0.0 were automatically
  carried over to 1.0.1, so I just added what I know won't be fixed.


RichardJJG:
Thanks!

PR #824, opened 2022-01-24
Add details about using Tap Initializer to generate yaml files for Runtime Resource Visibility

obabativavmware:


YugangW:
Thanks for the PR, @obabativavmware. Could you please kindly confirm if the changes you made are applicable to branch `1-0` (GA) as well? Thanks!

PR #827, opened 2022-01-24
Add install instructions for TBS for an air-gapped environment

tylerphelan:


tylerphelan:
setting as a draft for build service team members to vet before merging

matthewmcnew:
This looks great! I do wonder if we need the full airgapped install instructions here as the only thing unique about airgapped TBS (different from other packages) is that the deps-updater can be not be used and the deps must be installed via kp or imgpkg/kp (which is currently linked out of the docs). 
Could we just have a note on if being installed in an offline environment?

tylerphelan:
@matthewmcnew yea I hear that, I felt weird doing that when TAP didn't have any info on air-gapped installs. So maybe this PR should really be: "How to install TBS air-gapped assuming tap has already been relocated and the tap package-repo has been added"

tylerphelan:
See new PR which does not contain all of the TAP offline install steps: https://github.com/pivotal/docs-tap/pull/840

tylerphelan:
closing in favor of #840

PR #831, opened 2022-01-26
Update deployment_details.md

mhoshi-vm:
Not "Ready" but "Read"


mhoshi-vm:
It seems we need to fix the product configured default value also

PR #845, opened 2022-02-01
Add CNRS config so we have a domain for apps

drawsmcgraw:
This adds the CNRS config to the "full" and "light" profile `tap-values.yaml` config files. Without this, a `tanzu apps workload create` will fail because the app will use the default `example.com` TLD


a-peek4:
@sonjawriter @piachakra

PR #846, opened 2022-02-02
Rename `tekton` -> `tekton-pipelines`

chenbh:
This is more consistent with the package name in tap-packages which was a point of confusion. Also, since the package is part of the profiles, there's no need to install it separately

I don't think so?


RobC55:
@chenbh
Only pushed to docs-stage 1.1; not pushed to PROD 1.0 per https://github.com/pivotal/docs-tap/pull/846#issue-1122129595

PR #870, opened 2022-02-09
Update install-general.md

piachakra:


RichardJJG:
Thanks, @piachakra -- I'm guessing this needs cherry-picking to the `1-0` branch as well?

RichardJJG:
Actually, never mind -- I've read more from engineers about this section now and I know that while this info applies to v1.0, this particular PR does not need cherry-picking.

PR #874, opened 2022-02-09
fixed 1.0.0 to 1.0.1

krisapplegate:


RichardJJG:
Thanks, instead of a straight merge, we'll use this content to create a separate procedure for v1.0.1 so as not to mislead v1.0.0 users.

PR #881, opened 2022-02-11
Added GKE Regional Recomendation to Prereqs

krisapplegate:
Applies to 1.0-1.1+


RichardJJG:
Thanks, @krisapplegate!

PR #884, opened 2022-02-14
chore: Add upgrading info for scst-store

xtreme-shane-lattanzio:


RichardJJG:
Thanks, @xtreme-shane-lattanzio, does this content apply to TAP v1.0 as well as v1.1? We'll cherry-pick it after editing, if so.

xtreme-shane-lattanzio:
@RichardJJG This main applies to 1.1. That will be the first version with the new changes although some of the problems will still exist in 1.0. I will make another PR to clarify that. Thanks for checking and merging!

PR #894, opened 2022-02-18
Consistent version name

andreizimin:

TAP_VERSION variable is defined in the beginning of the document. To be used instead of VERSION


RichardJJG:
Thanks, @andreizimin! I tweaked your PR just a little more to make the placeholder and description match those earlier in the doc.

PR #895, opened 2022-02-18
Repaired TAP-GUI catalog links and wording

krisapplegate:


RichardJJG:
Thanks, @krisapplegate!

PR #898, opened 2022-02-22
Update release notes

chaomonica:
Which other branches should this be merged with (if any)? none (only 1-0)

### DIFF
--- a/release-notes.md
+++ b/release-notes.md
@@ -10,6 +10,11 @@ This topic contains release notes for Tanzu Application Platform v1.0.
 
 This release has the following known issues:
 
+#### Supply Chain Security Tools – Scan
+- **Scan Phase does not correctly reflect `Error`**: If an error (e.g. ScanPolicy defined or invalid xml) occurs during scanning, the phase remains as `Scanning`. The current workaround is to `kubectl describe` the scan and view the status and events.
+- **Two scan jobs/pods running at the same time**: There is an edge case where two scan jobs/pods appear when a scan policy is updated. This does not affect the result of the scan.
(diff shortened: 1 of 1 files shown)


RichardJJG:
Thanks!

RichardJJG:
A few questions:

1. What does "jobs/pods" mean? "jobs and pods"? "jobs or pods"? Is "jobs" just a synonym for "pods" here?

2. What is meant by the example error "ScanPolicy defined"? Is it an error that it is defined or not? Is ""ScanPolicy defined" the name of a common error? We need to delete this or add a bit more detail to this example.

3. "logs are missing error details with a `git clone` error or error fetching scanner logs"
I don't know what this means. I'm guessing it means:
"Running `git clone` sometimes causes errors that are only partially described in the event log. The same is true for errors arising from trying to fetch scanner logs."

chaomonica:
@RichardJJG If the below looks better, I can open another pull request to fix the wording.
1. It is two jobs and two jobs.  Currently instead of a single job and a single pod running, we are getting duplicates.
2. Oops, my mistake, it should be `ScanPolicy undefined`. 
3. "Errors that arise through either `git clone` or when trying to fetch scanner logs are only partially described in the event log."

RichardJJG:
> @RichardJJG If the below looks better, I can open another pull request to fix the wording.
> 
> 1. It is two jobs and two jobs.  Currently instead of a single job and a single pod running, we are getting duplicates.
> 2. Oops, my mistake, it should be `ScanPolicy undefined`.
> 3. "Errors that arise through either `git clone` or when trying to fetch scanner logs are only partially described in the event log."

Yes, please open another PR -- thanks!

PR #900, opened 2022-02-23
scst - Remove issues fixed in build 2

chaomonica:
Which other branches should this be merged with (if any)? only 1-0


RichardJJG:
...easy come, easy go XP
Thanks, @chaomonica

PR #901, opened 2022-02-23
Do not default enable_automatic_dependency_updates to true

tylerphelan:
- This field should be explicitly configured. Defaulting it to true can cause breaking changes.

1.0


dumez-k (approved):

RichardJJG:
Thanks! We'll change the `true/false` bit a little because we've had a complaint before that it confused someone. The value is `true` or `false`, after all, not `true/false`.

tylerphelan:
@RichardJJG can we chat in vmware slack? (me: @tyler TylerPhelan)

RichardJJG:
For anyone else reading, we chatted on Slack and resolved this. The final change will likely go live in under 24 hours.

PR #908, opened 2022-02-24
Update version of app live view to 1.0.2

vudayani-vmw:
This change is for TAP 1.0.2

### DIFF
--- a/tap-gui/plugins/install-application-live-view.md
+++ b/tap-gui/plugins/install-application-live-view.md
@@ -38,12 +38,12 @@ To install Application Live View:
     $ tanzu package available list run.appliveview.tanzu.vmware.com --namespace tap-install
     - Retrieving package versions for run.appliveview.tanzu.vmware.com...
       NAME                              VERSION        RELEASED-AT
-      run.appliveview.tanzu.vmware.com  1.0.2-build.2  2022-01-24T00:00:00Z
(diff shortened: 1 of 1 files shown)


RichardJJG:
@vudayani-vmw, in these commands `1.0.2` refers to the Tanzu Application Platform version the user is on, right? And so the number in the command needs to be, for example, `1.0.1` if the user is on Tanzu Application Platform v1.0.1?

vudayani-vmw:
@RichardJJG The version `1.0.2` here refers to Application Live View packages. The component package versions are independent of the Tanzu Application Platform version. For example, we had shipped App live View v1.0.1 for TAP 1.0.0 release.

RichardJJG:
@vudayani-vmw, okay, thanks, so the commands will work regardless of the TAP version the user is on (v1.0.0, v1.0.1, or v1.0.2)?

vudayani-vmw:
@RichardJJG , The App Live View  `1.0.2` is shipped with TAP v1.0.2. If the user is on TAP (v1.0.0, v1.0.1) they have to upgrade the TAP repository to 1.0.2 version so that new package versions are available and then upgrade the component packages.

RichardJJG:
Okay, @vudayani-vmw, we won't merge this until the TAP v1.0.2 release.

CC @a-peek4 and @piachakra

PR #910, opened 2022-02-24
Document pinned descriptor version tbs for TAP 1.0.2

tylerphelan:

1-0 so it can be in 1.0.2

This change should also be on the main branch because we will always have a pinned version, it will just need to be updated on each release of tbs


tylerphelan:
currently draft PR so tbs team can vet

tylerphelan:
good to go!

a-peek4:
cc: @piachakra merging this for the 1.0.2 release

PR #913, opened 2022-02-24
Added required namespace switch

parth-pandit:
All these commands need the developer namespace to be specified to make them work.


RichardJJG:
CC @pia for approval to merge this. `getting-started.md` is not, strictly speaking, an installation topic, but it's very similar.

piachakra:
@RichardJJG please merge

PR #920, opened 2022-03-02
Kdhaduk tap 1.0.2

dhadukk:

### DIFF
--- a/install-tanzu-cli.md
+++ b/install-tanzu-cli.md
@@ -249,7 +249,7 @@ the CLI core and plug-ins are installed:
 
     The expected output is:
 
-    * `version: v0.11.1` for Tanzu Application Platform v1.0.1
+    * `version: v0.11.1` for Tanzu Application Platform v1.0.1 or v1.0.2
     * `version: v0.10.0` for Tanzu Application Platform v1.0.0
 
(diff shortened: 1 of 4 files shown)


install-tanzu-cli.md
```diff
@@ -407,7 +520,7 @@ Follow these instructions to update the Tanzu CLI that was installed for a previ
 3. Navigate to [Tanzu Application Platform](https://network.tanzu.vmware.com/products/tanzu-application-platform/)
 on Tanzu Network.
 
-4. Click the **tanzu-cli-v0.11.1** directory.
+4. Click the **tanzu-cli-tap-1.0.2** directory.
```
mdc81, line 523: Has the name of cli bundle changed from v<version> to tap-<version>? Also, in all the versions mentioned, we seem to prefix "v", but here it is removed. Is it intentional?

RichardJJG:
@piachakra, we're seeing disagreement here. Please let us know when we can merge.

install-tanzu-cli.md
dhadukk: changed directory back to tanzu-cli-v0.11.1.

dhadukk:
@RichardJJG @piachakra This PR is ready to be merged.

PR #922, opened 2022-03-02
Update scst-store docs with tanzu insight plugin

julia-pu:
This is for TAP 1.1.0


RichardJJG:
Thanks, @julia-pu, very thorough!

PR #924, opened 2022-03-02
Update grype version for release

chaomonica:
Which other branches should this be merged with (if any)? Please cherry-pick to 1-0


RichardJJG:
Thanks, will do!

PR #925, opened 2022-03-02
Update resolved issues and removed duplicate known issue

chaomonica:
Which other branches should this be merged with (if any)? None


RichardJJG:
Hi, @chaomonica, should the known issue really be deleted for v1.0.1 here? If the fix is in v1.0.2 then the issue continues to exists for v1.0.1 users, right?

RichardJJG:
Thanks, @chaomonica!

PR #926, opened 2022-03-03
App live view install docs update for 1.1.0

vudayani-vmw:
These changes are only targeted for Tap 1.1.0


RichardJJG:
Thanks, @vudayani-vmw!

PR #929, opened 2022-03-03
[VSEE-459] Update scantemplate docs to show 3 scantemplates by default

rxl7906:
Which other branches should this be merged with (if any)? main

Update scantemplate docs to show 3 scantemplates by default (`public-source-scan-template`, `public-image-scan-template`, `blob-source-scan-template`)

If targetImagePullSecret is set in the tap-values.yml, then a `private-image-scan-template` would also be deployed.

If targetSourceSshSecret is set in the tap-values.yml, then a `private-image-scan-template` would also be deployed.

### DIFF
--- a/scst-scan/explanation.md
+++ b/scst-scan/explanation.md
@@ -36,7 +36,7 @@ By default, there will be 3 scan templates deployed (`public-source-scan-templat
 
 If targetImagePullSecret is set in the tap-values.yml, then a `private-image-scan-template` would also be deployed.
 
-If targetSourceSshSecret is set in the tap-values.yml, then a `private-image-scan-template` would also be dep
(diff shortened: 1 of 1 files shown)


rxl7906:
@RichardJJG sorry about that will do main branch next time.

Also can we cherry-pick this to main branch?

RichardJJG:
> @RichardJJG sorry about that will do main branch next time.
> 
> Also can we cherry-pick this to main branch?

Sure thing, a writer will cherry-pick when ready.

rxl7906:
@RichardJJG Thanks and sorry I forgot but this PR https://github.com/pivotal/docs-tap/pull/948 which is a typo fix for this PR needs to be cherry-picked to main also. Sorry!

PR #930, opened 2022-03-03
Update the new accelerator getting started content for 1.1

trisberg:

Only 1.1


RichardJJG:
Thanks, @trisberg!

PR #932, opened 2022-03-03
Added Multi-cluster View instructions for TAP-GUI

krisapplegate:
Applies to TAP 1.1


RichardJJG:
Thanks, @krisapplegate!

PR #933, opened 2022-03-03
Move resolved issue that resurfaced into known issues

chaomonica:
Which other branches should this be merged with (if any)? none


RichardJJG:
Thanks, @chaomonica!

PR #939, opened 2022-03-04
Added note on TAP GUI database and app_config

andreizimin:
For 1.1


RichardJJG:
Thanks, @andreizimin!

PR #940, opened 2022-03-04
Updated profiles

mpriamo:
Please promote to 1.1.0


RichardJJG:
Thanks, @mpriamo! This is a change to an installing topic, @piachakra, but my thinking is that we're safe to merge these when the changes are exclusive to v1.1 (i.e. far away from customers' eyes)...?

PR #942, opened 2022-03-04
updating plugin install instructions

msrinivasssa:

### DIFF
--- a/install-tanzu-cli.md
+++ b/install-tanzu-cli.md
@@ -330,6 +330,8 @@ To perform a clean installation of the Tanzu CLI plug-ins:
     ```
     cd $HOME/tanzu
     tanzu plugin install --local cli all
+    
+    tanzu plugin install --local cli/standalone all
     ```
 
 3. Check the plug-in installation status by running:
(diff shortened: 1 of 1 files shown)


RichardJJG:
@msrinivasssa, do these changes apply to v1.0 also?

CC @piachakra because I need go-ahead from her before we merge changes to installing topics that affect TAP v1.0.

msrinivasssa:
> @msrinivasssa, do these changes apply to v1.0 also?
> 
> CC @piachakra because I need go-ahead from her before we merge changes to installing topics that affect TAP v1.0.

No. This is applicable only for 1.1.0

RichardJJG:
Okay, thanks, @msrinivasssa -- I've merged. CC @piachakra.

a-peek4:
@msrinivasssa Can you clarify if the user runs both of the following commands:
```
tanzu plugin install --local cli all

tanzu plugin install --local cli/standalone all
```

Do these commands install a different CLI?

msrinivasssa:
> @msrinivasssa Can you clarify if the user runs both of the following commands:
> 
> ```
> tanzu plugin install --local cli all
> 
> tanzu plugin install --local cli/standalone all
> ```
> 
> Do these commands install a different CLI?

Yes user has to run both the clis. They install different plugins

PR #943, opened 2022-03-04
Update install.md

piachakra:
Also apply change to TAP 1.0.0 documentation (1-0) branch


RichardJJG:
Thanks, @piachakra. Is this part of a move to call components "packages" from now on? I've not noticed much inconsistency for this, per se, but we should have a rule, really.

PR #945, opened 2022-03-04
update app live view sections for 1.1

vudayani-vmw:
These changes are targeted only for the 1.1 release


RichardJJG:
Thanks, @vudayani-vmw!

PR #946, opened 2022-03-04
Clarify non-working use cases

jhvhs:
only in main


RichardJJG:
Thanks, @jhvhs!

YugangW:
Thanks for the PR, @jhvhs. Just wanted to double check are these changes intended to be applied to `main` branch as the notes about the unsupported use case for 1.0x will not be visible for users until we publish the `main` branch for 1.1 release in a few weeks. Thanks!

PR #948, opened 2022-03-04
Fix typo for scst-scan explanation page

rxl7906:
Which other branches should this be merged with (if any)? main

This is a follow-up from https://github.com/pivotal/docs-tap/pull/929. We were told to do changes on main but since this is a follow-up, I didn't want to create merge conflicts. But moving forward I will try to do main changes primarily.


RichardJJG:
Thanks for the contribution and explanation, @rxl7906!

PR #952, opened 2022-03-04
fixed typo

mpriamo:
1.1.0 change


RichardJJG:
Thanks
CC @piachakra because this is a change to an installing topic (albeit in `main` only)

PR #955, opened 2022-03-04
Added API Docs plugin to the TOC

andreizimin:
for 1.1


RichardJJG:
Thanks!

PR #957, opened 2022-03-04
Add Detailed Role Permissions Breakdown

Manifaust:


RichardJJG:
Thanks, @Manifaust, are all these changes just for the v1.1 docs?

Manifaust:
@RichardJJG sorry I forgot to type this in. Yes this is only meant for v1.1

PR #959, opened 2022-03-07
Separate claims from workload creation in STK usecases

edwardecook:
Update the services-toolkit getting started guide to reflect the separation of resource claims from workload creation.

N/A


RichardJJG:
Thanks, @edwardecook!

PR #960, opened 2022-03-07
docs: [TAA-2] Added instructions to install the auth plugin and fixed…

ohlookadollar:
… the cli command docs

This should only be for the 1.1 release


RichardJJG:
Thanks, @ohlookadollar!

PR #961, opened 2022-03-08
Remove line that is no longer needed

chaomonica:
Which other branches should this be merged with (if any)? Please cherry-pick to 1-0 after docs are no longer frozen.


RichardJJG:
Thanks, @chaomonica!

PR #965, opened 2022-03-10
Remove Light and Initial Multi-cluster docs

krisapplegate:
1.1+

### DIFF
--- a/install.md
+++ b/install.md
@@ -256,83 +256,7 @@ this can reuse the `tap-registry` secret created in
 
 ### <a id='light-profile'></a> Light Profile
 
-The following is the YAML file sample for the light-profile:
-
-```
-profile: light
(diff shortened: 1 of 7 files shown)


RichardJJG:
Are we okay to merge, @piachakra?

krisapplegate:
Confirmed with @mpriamo that this is good to merge now.

RichardJJG:
Thanks, merged!

YugangW:
Hi @krisapplegate, Thanks for the PR. I removed the `Reference` heading in `toc.md` as a temporary measure to fix the build failure. If you would prefer, an other way to fix this is to create a new file in the repo (e.g., `reference.md`) that the `Reference` heading can link to in the toc. I can create this blank file but might need your help to fill in some content if you prefer this method. Thanks!

PR #968, opened 2022-03-10
fix cli core *update* installation instructions

heyjcollins:
This change should be applied to TAP v1.0.0+


RichardJJG:
Are we okay to merge, @piachakra?

RichardJJG:
Obtained approval from Pia over Slack

PR #969, opened 2022-03-11
add supply-chain objects to k8s-reader role

jamieklassen:
Since the addition of the supply-chain plugin in 1.1.0, there are a bunch more resource types that tap-gui will be querying for. In order to avoid a bunch of 403 errors cluttering up the runtime resources page, it'd be a bit better if users gave more permissions to tap-gui's service account.

Which other branches should this be merged with (if any)? None, I think; this is brand-new in 1.1.0.

### DIFF
--- a/tap-gui/cluster-view-setup.md
+++ b/tap-gui/cluster-view-setup.md
@@ -38,31 +38,72 @@ kind: ClusterRole
 metadata:
   name: k8s-reader
 rules:
-  - apiGroups: ['']
-    resources: ['pods', 'services', 'configmaps']
-    verbs: ['get', 'watch', 'list']
-  - apiGroups: ['apps']
-    resources: ['deployments', 'replicasets']
-    verbs: ['get', 'watch', 'list']
-  - apiGroups: ['autoscaling']
-    resources: ['horizontalpodautoscalers']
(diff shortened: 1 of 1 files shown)


ryan-a-baker:
👍

YugangW:
Hi @jamieklassen, Thanks for the PR. I noticed all the indents were removed from line 41 to line 106. Just wanted to double confirm this change is intended. Here is the staging link for your review: https://docs-staging.vmware.com/en/Tanzu-Application-Platform/1.1/tap/GUID-tap-gui-cluster-view-setup.html Thanks!

jamieklassen:
@YugangW I thought it was nicer to use a consistent level of indentation for YAML arrays -- before there were some arrays that were indented and some that weren't, so I switched to none indented. But you've got me wondering -- is there a style guide/norm for these docs about that?

YugangW:
Thanks @jamieklassen for your quick response! Unfortunately, as far as I know, our current style guide does not apply to any code samples... In most cases, we rely on engineer's judgements when making any changes in the code samples. That's why we always double check with the engineers even when there is a slight doubt... I am really glad that all the changes you made are intended and the staging page looks good to you if it's not a wild assumption to make :-)

PR #976, opened 2022-03-16
tbs standalone descriptor name

tylerphelan:
- This user input should still be 1.0.0 even when running tap 1.0.1/2

This only needs to be merged into 1-0 branch


RichardJJG:
Thanks, @tylerphelan

PR #978, opened 2022-03-17
Update install.md to include reference to pre-reqs for container image registry requirements

mdc81:
adding reference to Prerequisites.


mdc81:
> Thanks, @mdc81, I'll copy this across to `main` as well given it applies to the v1.1 docs. For future reference, where possible we prefer that PRs are made to `main` only (where the v1.1 doc live).
> 
> The reason for this is that it's easier for the writers and engineers to deal with just one PR made to the most recent applicable docs version. Then the writers can easily copy across their edits to earlier branches and the engineer doesn't have to worry about the earlier docs versions.

sure, will do that from next changes.

PR #981, opened 2022-03-17
Update RBAC directions in namespace setup to explain service account and developer access to namespaces

evankanderson:

TAP 1.1 if already cut

### DIFF
--- a/install-components.md
+++ b/install-components.md
@@ -125,7 +125,7 @@ that you plan to create the `Workload` in:
     apiVersion: v1
     kind: ServiceAccount
     metadata:
-      name: supplychain
+      name: default
     secrets:
       - name: registry-credentials
     imagePullSecrets:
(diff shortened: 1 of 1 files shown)


install-components.md
```diff
@@ -127,71 +125,66 @@ that you plan to create the `Workload` in:
     apiVersion: v1
     kind: ServiceAccount
     metadata:
-      name: default
+      name: supplychain
```
nebhale, line 128: *issue*: `supplychain` isn't the right name for this SA as it's tied to a specific `Workload` (and a `Delivery` in this documentation).

nebhale (changes requested):

install-components.md
evankanderson: I can switch this back to `default` if we want; long-term it seems like we might want to separate "what identity does the supply chain run as" and "what identity does the deployed workload run as". With the proposed RoleBindings, we'd also need to add `ImagePullSecrets` for the "identity the deployed workload runs as", so it may be reasonable to separate these for now.

evankanderson: (Switched back to `default` for now)

RichardJJG:
Awaiting @piachakra's go ahead before merging because the PR affects an installation topic.

piachakra:
This is ready to merge @RichardJJG

PR #989, opened 2022-03-22
Document how to enable kapp-controller to operate on relocated images in a registry with a private CA Certificate

evankanderson:
Document how to load CA certificates into `kapp-controller`. With the switch to recommending image relocation, users with registries secured by a private CA will fail to install the package repository or packages due to certificate trust issues.

(Validated on `kind` with an on-cluster registry created by https://github.com/evankanderson/k8s-private-local-registry)

TAP 1.0, which also contains the image relocation instructions.

### DIFF
--- a/install-tanzu-cli.md
+++ b/install-tanzu-cli.md
@@ -103,7 +103,7 @@ Tanzu Application Platform v1.1 is supported on Cluster Essentials v1.0 and v1.1
 
     ```
     kubectl create namespace kapp-controller
-    ```\
+    ```
 
    Create a configuration secret using the registry's `ca.crt` stored on local disk:
 
--- a/install-tanzu-cli.md
+++ b/install-tanzu-cli.md
@@ -97,6 +97,22 @@ Tanzu Application Platform v1.1 is supported on Cluster Essentials v1.0 and v1.1
 
     - `DOWNLOADED-CLUSTER-ESSENTIALS-PACKAGE` is the name of the cluster essentials package you downloaded.
 
+1. **If your registry needs a custom certificate**, you will need to [load that configuration](https://carvel.dev/kapp-controller/docs/v0.32.0/controller-config/) into the cluster before installing `kapp-controller`. If your registry uses a public certificate, this step is not needed.
+
+   Create the `kapp-controller` namespace:
+
+    ```
+    kubectl create namespace kapp-controller
+    ```\
+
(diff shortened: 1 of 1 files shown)


YugangW (commented):
Thanks for the PR @evankanderson. Just to double check- is the backslash `\` in line 106 intended to appear outside of the code block? Thanks!

evankanderson:
Updated, thanks for the quick review!

install-tanzu-cli.md
```diff
@@ -97,6 +97,22 @@ Tanzu Application Platform v1.1 is supported on Cluster Essentials v1.0 and v1.1
 
     - `DOWNLOADED-CLUSTER-ESSENTIALS-PACKAGE` is the name of the cluster essentials package you downloaded.
 
+1. **If your registry needs a custom certificate**, you will need to [load that configuration](https://carvel.dev/kapp-controller/docs/v0.32.0/controller-config/) into the cluster before installing `kapp-controller`. If your registry uses a public certificate, this step is not needed.
```
evankanderson, line 100: Could use some style help here:

1. Is **bold** the right way to make this call out?
2. Should this be a positive statement (`If <condition>, do <step>`), negative statement (`unless <condition>, skip this this step. ...`), or both?

YugangW:
Merged. Thanks, @evankanderson!

install-tanzu-cli.md
YugangW: Thanks for pointing these out, @evankanderson.
- Re: comment 1
  - VMware style recommends using the following style for optional step like this (bold fonts, they are generally not recommended for non UI elements):
    - step #. (Optional) step descriptions (E.g., 1. (Optional) If your registry...)
    - I've already made these changes as part of my editing work. (Please let me know if the edits alter the original intents of the instructions.)
       - 1.1 staging: https://docs-staging.vmware.com/en/draft/Tanzu-Application-Platform/1.1/tap/GUID-install-tanzu-cli.html#install-cluster-essentials-for-tanzu-3
       - 1.0 staging: https://docs-staging.vmware.com/en/Tanzu-Application-Platform/1.0/tap/GUID-install-tanzu-cli.html#install-cluster-essentials-for-vmware-tanzu-3
- Re: comment 2
  - Both statements look good to me, but I personally prefer the positive statement a bit more as it's more straightforward and easier for the readers to follow.

PR #998, opened 2022-03-25
Update descriptor_name valid values

tylerphelan:
- This should only be full or lite

This should be in main and 1-1


RichardJJG:
Before we cherry-pick this to 1-0, @piachakra, can you confirm this is safe to merge? (It concerns an installation topic.)

PR #1001, opened 2022-03-28
scc: bump versions, update configurable props

cirocosta:
### proposed changes

- configurable properties: in the newer scc packages (ootb-*),
  `git_implementation` has been added to the supply chains and delivery,
  and `excluded_templates` to `ootb-templates

- version bump: the latest build of tap (1.1.0-build.14) includes new
  versions (those which are bumped to in this commit)

### which other branches should this be merged with (if any)?

these are changes that have been made aiming at 1.1.0+.


scc/install-ootb-sc-basic.md
```diff
@@ -43,6 +43,9 @@ To install Out of the Box Supply Chain Basic:
                               be pushed to (required).
 
 
+		git_implementation        Determines which git client library to use.
```
cirocosta, line 46: whops, will fix the indentation

PR #1011, opened 2022-03-30
Add release notes for app accelerator

trisberg:

This is for TAP 1.1.0


a-peek4:
Thanks! @trisberg

PR #1013, opened 2022-03-30
fix: typo and name of RO metadata store service account to match gett…

rawlingsj:
…ing token command


rawlingsj:
specifically this fixes the name of the service account so that it matches the service account referenced in this step https://github.com/pivotal/docs-tap/blob/ee2577c7e3f363b5fbba149f5e1961222097e0e2/scst-store/create-service-account-access-token.md?plain=1#L126

a-peek4:
Thanks! @rawlingsj

PR #1017, opened 2022-03-31
Services Toolkit 0.6.0 updates

teddyking:
Hi there,

This commit includes a complete rewrite of "Section 5: Consuming
Services on Tanzu Application Platform" of the Getting Started Guide.
Prior to this commit, this section of the docs was very focussed on the
"Services Toolkit" component and as such did not provide an appropriate
level of detail for users of Tanzu Application Platform. The new content
in this commit should be much more relevant.

We have also done away with the "Use case" based approach, opting
instead to provide 1 detailed walkthrough for the most common use case -
binding Application Workloads to Service Instances. We have linked off
to the Services Toolkit component documentation for additional, more
advanced use cases so as not to clutter the Getting Started Guide.

Finally, this commit bumps version numbers of Services Toolkit from
0.5.0 to 0.6.0.

Which other branches should this be merged with (if any)? None. These changes are for TAP 1.1.


a-peek4:
Thanks @teddyking 
cc: @sonjawriter

PR #1021, opened 2022-03-31
Reworked query.md and add.md based on feedback

miketeve:
Reworked query-data page, fixed "insight create" to "insight add", removed need for "Additional Resources" page, clarified "source commit" throughout


a-peek4:
Thanks @miketeve

PR #1023, opened 2022-03-31
add source controller and apps plugin changes

heyjcollins:
DOCS TEAM - I added some links to the release notes --- can you please doublecheck and make appropriate fixes so they will work as expected when rendered in the docs site?


RichardJJG:
Thanks, @heyjcollins, will do

PR #1026, opened 2022-04-01
add checkmark to view profile for tap-auth/default roles

dherbrich:
Which other branches should this be merged with (if any)? all branches for TAP 1.1+

this checkmark needs to be set because tap-auth is also in view profile in TAP 1.1. it was decided to not change this after code freeze so we will change it (and also change the docs) for version 1.2.
Cc: @Manifaust


a-peek4:
Thank you @dherbrich

PR #1029, opened 2022-04-01
Update release-notes.md

piachakra:
Which other branches should this be merged with (if any)? We will need this added to all the releases going forward so will need to merge to main as well

### DIFF
--- a/release-notes.md
+++ b/release-notes.md
@@ -2,6 +2,44 @@
 
 This topic contains release notes for Tanzu Application Platform v1.0.
 
+## <a id='1-0-2'></a> v1.0.3
+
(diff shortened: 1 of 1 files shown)


RichardJJG:
@piachakra, we copy across known issues to the release notes in subsequent minors (if they still apply) but not resolved issues. (If something was fixed in an earlier patch, minor, or major, the reader can assume the fix is present in the new patch/minor/major unless told otherwise.)
 
I've compared the release notes in `main` to this PR and seen nothing to carry across: the two known issues that still apply are already there and the third known issue shows up in the **Resolved issues** section. 

If I've missed something, just let me know.

PR #1034, opened 2022-04-01
Update functions.md

sheltoG:
Fixed YAML indentation.


RichardJJG:
Hi, @sheltoG, the original indentation is needed so as not to break the list formatting that the snippets sit in. Thanks for the effort, though.

PR #1038, opened 2022-04-04
additional copy for management port update

heyjcollins:

this should be applied to TAP v1.1+ docs

### DIFF
--- a/spring-boot-conventions/reference/CONVENTIONS.md
+++ b/spring-boot-conventions/reference/CONVENTIONS.md
@@ -44,7 +44,7 @@ If desired, the JAVA_TOOL_OPTIONS value(s) can be provided in the workload.yaml 
     spec:
      env:
      - name: JAVA_TOOL_OPTIONS
-      value: -Dmanagement.server.port=8082
+       value: -Dmanagement.server.port=8082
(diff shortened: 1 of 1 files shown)


spring-boot-conventions/reference/CONVENTIONS.md
```diff
@@ -17,6 +19,37 @@ The submitted pod intent can be:
...
+1. Set property using `workload.yaml`:<br/>
+If desired, the JAVA_TOOL_OPTIONS value(s) can be provided in the workload.yaml as follows:
+    ```
+    apiVersion: carto.run/v1alpha1
+    kind: Workload
+    ...
+    spec:
+     env:
+     - name: JAVA_TOOL_OPTIONS
+      value: -Dmanagement.server.port=8082
```
atmandhol, line 47: can we add 1 more space here to be on the same level as `name`?

heyjcollins: good catch! - 
yes the "v" in "value" should be aligned with the "n" in "- name:" in the line above.

a-peek4:
Thanks @heyjcollins

PR #1039, opened 2022-04-04
update spring boot conventions notes

heyjcollins:
none


snneji:
Thanks @heyjcollins

PR #1040, opened 2022-04-04
Moved all insight docs to the cli-plugins folder

miketeve:
...and updated links

Do NOT merge with other branches; 1.1 only


RichardJJG:
Thanks, @miketeve!

PR #1053, opened 2022-04-06
Updating Cluster Essentials install instructions to point to CE docs …

rnandi:
…site

none


RichardJJG:
Thanks, @rnandi!

PR #1054, opened 2022-04-07
Fixed nesting issues plus page title tweaks

miketeve:


RichardJJG:
Thanks, @miketeve!

RichardJJG:
I've added "the" back where grammatically necessary. We can say just "Tanzu Insight" (pronoun) but need "the" for "Tanzu Insight plug-in" (common noun).

PR #1066, opened 2022-04-07
Update functions.md

sheltoG:
Fixed indentation of YAML code samples


a-peek4:
Hi @sheltoG These updates to the indentations will actually break your numbered list, so you should remove the indentation related changes before we merge this PR if you want to retain the numbered list

sheltoG:
Is there a way to preserve the indentations? The code wont work without them. I know we can do them looking at the TAP docs. I'm just not well versed in markdown :)

a-peek4:
@sheltoG Yes, you can have whatever indentations you want within the code block, just indent the entire code block itself as it was before in the original text before your PR (1 tab or 4 spaces in)

PR #1071, opened 2022-04-08
Remove enable_auto_updates config from main install examples

tylerphelan:
- This field defaults to false and has been a source of confusion. We want to remove
  it from the main install examples/flows and only make information about it
  available in more specific use-cases.
- Also remove the requirement to accept the TBS EULA which is not required.

This should be merged into main and 1-1 for the TAP 1.1 docs and for future versions.


matthewmcnew (approved):

tylerphelan:
Ready!

a-peek4:
Thanks @tylerphelan

PR #1072, opened 2022-04-08
Update functions.md

sheltoG:
Updated title to highlight beta feature and added more context to beta clause.


a-peek4:
@sheltoG Thanks, for future reference please refrain from editing writing in a way that tech writers will have to edit to be complete sentences again (like adding dashes in the middle of sentences). Thank you!

PR #1076, opened 2022-04-08
Removed specific versions from upgrades

krisapplegate:
1.1


a-peek4:
Thanks @krisapplegate

PR #1077, opened 2022-04-08
content/consistency updates

heyjcollins:
# Hi Docs Team! 

These changes should be applied to **TAP v1.1+**

A couple things:
1. I removed the comments on spec and env where someone wrote ":" and ":"
because if someone used specifications and/or environment instead of spec and/or env, the workload.yaml would fail.
so we CAN'T prefer those alternatives under any circumstances.
2. I added content informing users how to override the conventions


a-peek4:
Thanks @heyjcollins

PR #1078, opened 2022-04-08
Removed SCA scan results, SCA scan summary...

miketeve:
... and store-specific known issues

Which other branches should this be merged with (if any)? 1.0


a-peek4:
Hi @miketeve Can you confirm if the changes in the red box are the ones you want to keep with your commit?
Screen Shot 2022-04-08 at 2 53 24 PM

miketeve:
@a-peek4 i'm not super great at reading that.  if that is just showing that those files are deleted then, yes, that's what we want!

a-peek4:
Great, thanks. Yes that is what that shows.

a-peek4:
Merging bc it came in before docs freeze

PR #1079, opened 2022-04-08
Clean Up GitOps Supply Chain documentation

ryan-a-baker:
Doesn't make the document perfect, but cleans it up to the point that it should be able to be followed without any missing steps or errors.

Summarizing the changes: 

- Cleaned up introduction to explain what gitops method provides more concisely
- Provided an example of what the ssh-key looks like (didn't put in known hosts as I guess that could change and I'm honestly not sure if it's required, I added what I thought was correct and it just kept updating for me automatically because I did it wrong)
- Added directions to add the secret to the service account
- Correct the example repsitory_prefix to use SSH instead of Git
- Corrected values to have dot notation instead of underscore to be consistent

### DIFF
--- a/scc/ootb-supply-chain-basic.md
+++ b/scc/ootb-supply-chain-basic.md
@@ -417,6 +417,8 @@ of the supply chain having the configuration that got created by it pushed to a
 git repository where that is persisted and used at the basis for further
 deployments.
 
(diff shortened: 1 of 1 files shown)


a-peek4:
THanks @ryan-a-baker

PR #1083, opened 2022-04-08
scc: refactor gitops-related content

cirocosta:
### proposed changes

here I propose modifying the structure of the documentation that
relates to gitops in the following manner:

```
/scc

  /git-auth			details about how to create the k8s secrets,
				key pairs, etc related to git auth referenced
				in multiple places

  /building-from-source		focus on how one can build from source
				code that comes either from a git
				repository or local directory

  /gitops-vs-regops		focus on how one might publish the
				configuration produced by the supply
				chain to either a git repository or an
				image registry

  /pre-built-image		talk about the new capability of
				bringing your own container image

				(see `scc:pre-built-image`)
```

this way, we can speak about the details of each different twist that
can be performed w/ regards to the supply chains depending on certain
workload config / tap installations properties.

The changes here aim at TAP 1.1+

### DIFF
--- a/scc/building-from-source.md
+++ b/scc/building-from-source.md
@@ -1,0 +1,454 @@
+# Building from source
+
+Regardless of the Out of the Box Supply Chain Package installed, when it comes
+to providing source code for the Workload, that can either come from a
+developer's machine (directory in the filesystem) or a Git repository.
+
+```
+Supply Chain
+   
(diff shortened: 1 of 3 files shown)


a-peek4:
Hi @cirocosta Can you clarify if these are new features for 1.1 or a general proposal for new docs structure?

PR #1084, opened 2022-04-08
scc: bump versions to final versions

cirocosta:
### proposed changes

Bump the versions of the SCC component to the final ones:

- carto: 0.3.0
- ootb:  0.7.0

The changes here are specific to TAP v1.1+


a-peek4:
Thanks @cirocosta

PR #1085, opened 2022-04-08
scc: update ootb-scc-*

cirocosta:
### proposed changes

- authn/authz: make use of the `workload` and `deliverable` ClusterRoles
  instead of providing the full Role

- overview: mention the dual capabilities of the ootb-sc packages (from
  source vs pre-built images)

- basic: update the namespace setup to account for the placeholder
  secret for `tap-registry` secret as, without it, one wouldn't be able
  to have kpack/image etc running as they'd point at a serviceaccount
  that wouldn't provide pull secrets for tanzunet

note that we're able to remove a bunch of information about gitops
because those sections are now covered under #1083 

the changes here are for TAP v1.1+

### DIFF
--- a/scc/ootb-supply-chain-basic.md
+++ b/scc/ootb-supply-chain-basic.md
@@ -1,119 +1,130 @@
 # Out of the Box Supply Chain Basic
 
-This Cartographer Supply Chain ties together a series of Kubernetes resources which
-drive a developer-provided Workload from source code to a Kubernetes configuration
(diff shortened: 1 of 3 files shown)


a-peek4:
Hi @cirocosta I've got a few merge conflicts that I need your help resolving. Can you reach out to #tap-docs to connect with someone on my team on Monday about this PR? Thanks

PR #1089, opened 2022-04-12
fix: typo and name of RO metadata store service account to match gett…

rawlingsj:
…ing token command


RichardJJG:
Thanks!

PR #1091, opened 2022-04-12
declare CVE fix in TAP GUI

jamieklassen:
[PSTAR-592](https://jira.eng.vmware.com/browse/PSTAR-592) for context

Which other branches should this be merged with (if any)? I'll open a separate PR to disclose this CVE in past versions


RichardJJG:
Thanks, @jamieklassen!

PR #1095, opened 2022-04-13
Note namespace exclusion for scst-sign

DennyHoang:
Signed-off-by: Denny Hoang <dhoang@vmware.com>


HenryBorys:
@DennyHoang Is this PR for 1.2 only or also 1.1?

PR #1099, opened 2022-04-14
fix typo in run cluster explanation

jamieklassen:
Which other branches should this be merged with (if any)? I think just 1.1.0 is good, which is when we introduced multi-cluster docs.


RichardJJG:
Thanks!

PR #1100, opened 2022-04-14
fix reference to run cluster values-file key

jamieklassen:
Which other branches should this be merged with (if any)? I think just 1.1.0 is good, which is when we introduced multi-cluster docs.


RichardJJG:
Thanks!

PR #1105, opened 2022-04-18
Update query-data.md

doddatpivotal:
Expected query parameters were not valid for the second sourcescan example.  Fixed.


YugangW:
Thanks for the PR @doddatpivotal! Do you also want to apply these changes to `1-1` and `1-0`?

doddatpivotal:
Yes, that make sense.

PR #1106, opened 2022-04-18
Update functions.md

sheltoG:
Updated code samples and Tanzunet image URL.

### DIFF
--- a/workloads/functions.md
+++ b/workloads/functions.md
@@ -29,8 +29,8 @@ To use the function `buildpacks`, you must upload their buildpackages to Build S
 
     ```
     kp clusterstore add default \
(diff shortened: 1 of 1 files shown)


RichardJJG:
Thanks, @sheltoG, before I merge:
1. `true` is a code value, so the code tags belong there.
2. Should I cherry-pick this to the `1-1` branch? (I'm guessing that this change applies to v1.1.0 as well as v1.1.1)?
3. Is [line 80](https://github.com/pivotal/docs-tap/pull/1106/files#diff-9b2065b4a55897e9121064cc3f5829886f6e22f85960c12c241b0f37555858a4R80) really valid? Readers are liable to copy and paste without thinking unless we explicitly show what needs editing in an example code snippet.

sheltoG:
3. Yeah we can cherry-pick this to the 1-1 branch.
4.  I agree, we can get rid of line 80 all together.

PR #1107, opened 2022-04-19
Update tap-values-build-sample.md

doddatpivotal:
Since grype is referenced, adds the appropriate settings for testing_scanning supply chain with reference to metadata store on the run cluster.

Which other branches should this be merged with (if any)? Should also be merged with 1.1


RichardJJG:
Thanks, @doddatpivotal!

PR #1110, opened 2022-04-21
Update default descriptor version for tbs to 100.0.293

tylerphelan:

This should be merged into `main` and `1-1` for the tap 1.1.x release version docs.


RichardJJG:
Thanks, @tylerphelan!

PR #1111, opened 2022-04-22
consolidate cli/plugin install and simplify copy

heyjcollins:
and improve readability of overall page with links to sections and some explanations.

### Should ONLY be applied to in TAP v1.1.0+ documentation


RichardJJG:
Thanks, @heyjcollins!

PR #1119, opened 2022-04-27
Update release notes for scan-controller and grype-scanner 1.1.1-build.1

chaomonica:
Which other branches should this be merged with (if any)? only 1-1


RichardJJG:
Thanks, @chaomonica!

PR #1120, opened 2022-04-28
fixing broken code block

kevinvmware:
This is the section "Set up developer namespaces to use installed packages", in the 3rd bullet point

Code block has incorrect yaml.

Please have engineer review (I'm not a yaml or Kubernetes expert)

### DIFF
--- a/install-components.md
+++ b/install-components.md
@@ -167,59 +167,33 @@ that you plan to create the `Workload` in:
 
     ```console
     cat <<EOF | kubectl -n YOUR_NAMESPACE apply -f -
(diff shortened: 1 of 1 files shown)


snneji:
Hi @kevinvmware, as you noticed, an engineer submitted PR #1122. I ran text compare on your PR vs PR #1122 and it confirmed that the engineer's PR makes the same changes that you recommended as well as adding some more details. (The only other differences are that the other PR changes the spacing around the whole code block, but that is a formatting concern and doesn't affect the code, and also uses placeholders where users should input their values). 

Because of that I'm going to close this PR. Thank you for your feedback!

PR #1121, opened 2022-04-28
Added section for 1.1.1 and move updates intended for 1.1.1 from 1-1

chaomonica:
Which other branches should this be merged with (if any)? 1-1


RichardJJG:
Thanks!

PR #1122, opened 2022-04-28
docs: [TANZUSC-1163] Fixed the rolebinding yaml and added more details about the binding

ohlookadollar:
1.1
1.2


RichardJJG:
Thanks, @ohlookadollar!

PR #1123, opened 2022-04-28
Add tap gui release notes and known issues for 1.1.1

austina-lin:


RichardJJG:
Thanks, @austina-lin!

PR #1139, opened 2022-05-10
fix TAP GUI values-file snippets

jamieklassen:
[PSTAR-609](https://jira.eng.vmware.com/browse/PSTAR-609) for some (scant) context.

@RichardJJG in our Slack discussion I mentioned some confusion around comments in these snippets, but less address that separately (if ever) -- correct indentation is much more urgent.

Which other branches should this be merged with (if any)? This will be relevant for 1.2 as well!

### DIFF
--- a/tap-gui/database.md
+++ b/tap-gui/database.md
@@ -14,18 +14,18 @@ To use a PostgreSQL database:
 
     ```yaml
         backend:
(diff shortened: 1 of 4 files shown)


RichardJJG:
Thanks again!

PR #1141, opened 2022-05-10
provide separate untar directions for linux vs. mac

heyjcollins:


RichardJJG:
Thanks, I'll cherry-pick backwards

PR #1143, opened 2022-05-10
improve setting path on Windows (step 9)

heyjcollins:
And added inline link to Windows install instructions for convenience and discoverability


RichardJJG:
Thanks, @heyjcollins, we can cherry-pick this to `1-1`, right?

YugangW:
Hi @heyjcollins, Just follow up with Richard's question- Could you please confirm if this change is applicable to `1-1`? Thanks!

heyjcollins:
Hey @YugangW and @RichardJJG - my answer is a Yes AND...

The change should be cherrypicked to 1.1

the change on line 187 (of my commit) can be applied to any/all docs versions where the Windows installation instructions include the original line that was updated --- which might be all the way back into the betas but it may only be worth going back to TAP v1.1.0+

PR #1144, opened 2022-05-11
App Live View doc integration changes

ksankaranara-vmw:


HenryBorys:
@ksankaranara-vmw Did you see that late yesterday I copied all of the ALV docs into the TAP repo? Given that, is this PR still needed?

HenryBorys:
@ksankaranara-vmw Oh, I see, you are making some changes.

HenryBorys:
1. @ksankaranara-vmw Just a minor point so we don't work at cross purposes when it comes to using the gerund vs. imperative verb form for certain titles. See https://confluence.eng.vmware.com/display/IXCS/Writing+chapter+titles,+headings+and+topic+titles. So for instance a topic that gives specific instructions to install x should be titled "Install x". Whereas a topic that discusses all kinds of other information about installing x but doesn't actually give the specific instructions, should be titled "Installing x". Most of our components get this right in the documentation, but TAP itself isn't the best example. So I've corrected this for Install ALV and Uninstall ALV.

ksankaranara-vmw:
> 1. @ksankaranara-vmw Just a minor point so we don't work at cross purposes when it comes to using the gerund vs. imperative verb form for certain titles. See https://confluence.eng.vmware.com/display/IXCS/Writing+chapter+titles,+headings+and+topic+titles. So for instance a topic that gives specific instructions to install x should be titled "Install x". Whereas a topic that discusses all kinds of other information about installing x but doesn't actually give the specific instructions, should be titled "Installing x". Most of our components get this right in the documentation, but TAP itself isn't the best example. So I've corrected this for Install ALV and Uninstall ALV.

Oh okay @HenryBorys makes sense. will keep that in mind

PR #1145, opened 2022-05-11
Update functions.md

sheltoG:
Added indents to optional lines.


a-peek4:
Thanks @sheltoG

PR #1146, opened 2022-05-12
Fix links in App Live View page

ksankaranara-vmw:


a-peek4:
Thanks @ksankaranara-vmw

HenryBorys:
@ksankaranara-vmw Thanks! That was on my list to do today. I'll check for other broken links just in case.

PR #1147, opened 2022-05-12
Add new workload samples for other application types.

evankanderson:
This needs editorial review. I think @snneji or @a-peek4 are probably the experts here.

Sorry about the huge chunks of in-line YAML; the goal was to have _something_ that customers can experiment with today in beta form while we work to make these capabilities native.

I'm not sure what else we might need, or where I've diverged from the underlying style, so please correct me where needed.

# Which other branches should this be merged with (if any)?

TAP 1.1


a-peek4:
Thanks @evankanderson

PR #1149, opened 2022-05-13
Fix ToC in App Live View

ksankaranara-vmw:


a-peek4:
Thanks @ksankaranara-vmw

PR #1150, opened 2022-05-14
Clarify how adding SecretExport works

Mpluya:
... by providing info on the SecretImport counterpart installed with the scan package


RichardJJG:
Thanks, @Mpluya, does this change apply to the v1.1 TAP docs also? (The main branch is for v1.2 docs)

Mpluya:
@RichardJJG yes, it does apply for v1.1. TAP docs. Let me know if I need to open this PR elsewhere. Thank you!

RichardJJG:
@Mpluya, no, that's okay, thanks. We always want engineers to PR to `main` only and then say in the opening comment whether the writer needs to cherry-pick the change to earlier branches (and which earlier branches, if so).

PR #1151, opened 2022-05-16
Update ingress-multicluster.md

x95castle1:
The yaml example provided in the "The Install Supply Chain Security Tools - Scan with the following configuration" section is missing "scanning:" as the top level property. This can cause someone to not realize the yaml is incorrect. I would be helpful to include the top level qualifier in the example so it matches what should be in TAP Values. 

This should be included in TAP 1.1 documents.


RichardJJG:
Thanks, @x95castle1!

PR #1154, opened 2022-05-18
update windows CLI and plugin install

heyjcollins:
based on going through the process, the instructions were a little bit off.

**Please cherrypick these windows-related updates back to TAP 1.0.0 and above** 

cc @BarrAndrea


YugangW:
Thanks @heyjcollins

PR #1160, opened 2022-05-20
docs: remove not supporting windows for insight plugin

julia-pu:
TAP 1.2


RichardJJG:
Thanks, @julia-pu

PR #1163, opened 2022-05-23
TAP-GUI Customization Docs

krisapplegate:
Applies to TAP 1.2+ Only


RichardJJG:
Thanks!

andreizimin:
Thank you, Kris and Richard!

PR #1168, opened 2022-05-24
Add RRV docs for pod logs

menehune23:
Targeting 1.2 docs


RichardJJG:
Thanks, @menehune23, these changes only apply to TAP v1.2 onwards, right?

menehune23:
@RichardJJG Yep! 1.2

PR #1171, opened 2022-05-25
Updated the header and 'New Portal Name'

andreizimin:

### DIFF
--- a/tap-gui/customize/customize-branding.md
+++ b/tap-gui/customize/customize-branding.md
@@ -1,4 +1,4 @@
-# Customizing branding
+# Customizing the Branding
 
 This topic describes how to customize the branding within the Tanzu Application Platform GUI portal.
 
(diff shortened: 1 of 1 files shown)


RichardJJG:
Thanks for the effort, @andreizimin, but topic titles are supposed to be sentence-case:
https://confluence.eng.vmware.com/display/IXCS/Capitalization?focusedCommentId=1322260427#comment-1322260427

And placeholders within code snippets must be uppercase with hyphens instead of spaces:
https://docs-wiki.tas.vmware.com/wiki/style/text-formatting-conventions.html#variables

VMware has literally thousands of style rules and they're listed across multiple websites. And the style rules sometimes change and only the writers are kept in the loop.

So...if you see something that you think is wrong stylistically, please do feel free to raise a PR (you might be right!) but know that non-writers are not expected to fix style errors.

RichardJJG:
As for "the branding" vs "branding"...either is fine in these cases.

RichardJJG:
Must the `PORTAL-NAME` value be title case with whitespaces for spaces? (I'm guessing that's why you changed the placeholder text -- to match the style the user is supposed to use?)

PR #1172, opened 2022-05-25
Update functions.md

sheltoG:
Added Functions CLI commands


snneji:
Thanks @sheltoG. Which TAP versions ist this for?

sheltoG:
Good question, this is for TAP 1.1 and 1.2. 

Thank you!

PR #1173, opened 2022-05-26
Add Documentation for Learning Center Plugin on TAP-GUI

rvallejony:


RichardJJG:
Thanks, @rvallejony! Do any of these changes apply to TAP v1.1 (I can cherry-pick, if so).

PR #1174, opened 2022-05-26
Add App Live View 1.1.2 docs

ksankaranara-vmw:


RichardJJG:
Thanks, @ksankaranara-vmw. I should move this content to a new v1.1.2 section within the release notes, right? (The content is in the v1.1 section, which contradicts the PR title.)

PR #1175, opened 2022-05-26
Create Iterate on your function

sheltoG:
Would it be possible to make the functions beta section a folder and add this file as a drop down item within it?

### DIFF
--- a/workloads/Iterate on your function
+++ b/workloads/Iterate on your function
@@ -1,0 +1,49 @@
+# Iterate on your function
+
+The VSCode extension enables live updates of your application while running on the cluster and allows you to debug your application directly on the cluster. 
+
+## <a id="prereqs"></a> Prerequisites
+    - Tanzu Dev Tools for Visual Studio Code
+    - Tilt v0.27.2 or later.
+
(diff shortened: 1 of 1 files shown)


RichardJJG:
@sheltoG, yes, we can nest topics. Is this content for just TAP v1.2 onwards?

sheltoG:
We can nest and add this content for TAP 1.1 docs as well. Thank you!

sheltoG:
My apologies, I was out last week. 

1.  We can use "Configure the Tanzu Developer Tools extension".
2. Step 3 is a note that may effect some users.
3. At the bottom of the VS code window there is text that will either say Live Update Started or Live update stopped.
4. The option they would select would be Tanzu: Live Update Stop.

snneji:
Hi, just a quick followup question in reference to your answer to question 4: would they also have the option "Tanzu: Live Update Continue" in the command palette?

sheltoG:
You only have two commands to choose from: Tanzu: Live Update Start and Tanzu: Live Update Stop. There is a section of the docs this is referenced from: [(https://docs.vmware.com/en/VMware-Tanzu-Application-Platform/1.1/tap/GUID-getting-started.html#live-update-your-application-7)

PR #1179, opened 2022-05-27
Document live hover feature

mingxiao:


RichardJJG:
Thanks, @mingxiao! Do these changes apply to TAP v1.1 as well (do want me to cherry-pick these changes to the 1-1 branch?)

mingxiao:
@RichardJJG no

kdvolder:
@RichardJJG now that this is merged were (if anywhere?) can we find these new docs in the staging area? I'd like to take another look at it in that format; if that is possible.

I was looking here: https://docs-staging.vmware.com/en/VMware-Tanzu-Application-Platform/index.html

But I only find in there v1.1 and v1.0 docs, whereas I think this will be part of 1.2 (or maybe I'm confused about all the versions :-). Anyhow if someone could give me a link to corresponding staging area, that would be swell.

RichardJJG:
@kdvolder, sure, the v1.2 docs are staged here:
https://docs-staging.vmware.com/en/draft/Tanzu-Application-Platform/1.2/tap/GUID-vscode-extension-about.html

We record where different branches are staged in the tap-docs repo [README](https://github.com/pivotal/docs-tap#readme).

kdvolder:
I tried getting to the doc section about "live hover feature" from that link, but I can't seem to get to it.
Perhaps this is deliberate (as the feature is deemed 'experimental' and it is disabled behind a 'feature flag').

However, it means still haven't been able to actually view the documentation live and I wonder if it is possible for anyone to get to it. Maybe it is only accessible by a direct url/link, and this might be intentional. But what is that 'magic' url? Even if it isn't adevertised publically, I'd like to put a pointer to it from [TAP-testing confluence page](https://confluence.eng.vmware.com/display/CNA/TAP+1.2.0+Bug+Bash).

kdvolder:
I tried guessing the url by 'reverse-engineering' the pattern of other urls. This is what I came up with:

https://docs-staging.vmware.com/en/draft/Tanzu-Application-Platform/1.2/tap/GUID-vscode-extension-live-hover.html

But its 404.

kdvolder:
@RichardJJG Sorry forgot to tag for attention ^^^

RichardJJG:
Ah, I see the problem, @kdvolder -- this PR doesn't include a link to the new topic in toc.md. 
I've just rectified that, so you can now see the topic now: 
https://docs-staging.vmware.com/en/draft/Tanzu-Application-Platform/1.2/tap/GUID-vscode-extension-live-hover.html

It still looks a bit rough -- no-one has edited it yet. The TAP v1.2 release isn't for a while yet, after all.

PR #1185, opened 2022-05-30
Add some detailed steps to live-hover docs

kdvolder:
Follow up from https://github.com/pivotal/docs-tap/pull/1179

I put myself into the shoes of  user reading the docs, and I felt like it might help to provide a concrete example/steps to actually see something working.

### DIFF
--- a/vscode-extension/live-hover.md
+++ b/vscode-extension/live-hover.md
@@ -11,3 +11,77 @@ In order to activate live hover, you need to start your vscode instance with `TA
 `> TAP_LIVE_HOVER=true code /path/to/project`
 
 Once you have a workload deployed, you should see the live hovers.
+
+## A Concrete Example
+
+Assuming you have the pre-requisited installed. Let's go through the 
+steps deploy the sample app [tanzu-java-web-app](https://github.com/sample-accelerators/tanzu-java-web-app) to our cluster; and get live
+hovers working.
+
+*Step 1:* Clone the repo 
+
+```
+$ git clone https://github.com/sample-accelerators/tanzu-java-web-app
+```
+
+*Step 2*: Open the project in vscode, with the live-hover feature enabled
+
+```
+$ TAP_LIVE_HOVER=true code ./tanzu-java-web-app
+```
+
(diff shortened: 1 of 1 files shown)


RichardJJG:
Thanks, @kdvolder! Does this change apply to TAP v1.1 as well? (Should I cherry-pick this change to the 1-1 branch?)

mingxiao:
@RichardJJG no, this does not apply to v1.1

RichardJJG:
@kdvolder, should there be steps to cover what to do if the user finds the wrong cluster is targeted? 
I'm thinking of [line 40](https://github.com/pivotal/docs-tap/pull/1185/files#diff-a7ecc79275f5c6b55246da00051925f58479f6b6116a85184bef3fc87d499bf2R40), specifically.

RichardJJG:
@kdvolder, should there be steps to cover what to do if the user finds the wrong cluster is targeted? 
I'm thinking of [line 40](https://github.com/pivotal/docs-tap/pull/1185/files#diff-a7ecc79275f5c6b55246da00051925f58479f6b6116a85184bef3fc87d499bf2R40), specifically.

kdvolder:
@RichardJJG I'm not sure what we would put. The steps for targetting a cluster with kubectl are pretty standard and I would assume that if people are using k8s and kubectl they would know how to point kubectl at a particular cluster (they would probably need to do that also just to deploy/run their workload in the first place).

There is however some configuration that can be done in vscode-tanzu-tools that someone could do the override the default / current kubectl context in vscode. But @mingxiao would know better how to do that and whether or not it is something that would best be explained here.

kdvolder:
> The steps for targetting a cluster with kubectl are pretty standard
Actually, in a way that isn't true. How you get 'kubectl access' to cluster varies greatly depending on how a cluster was provisioned. But the gist of it is that we have to assume someone able to run a workload on a cluster must have figured out those steps for their particular environment already (or else they cannot deploy a workload).

RichardJJG:
Okay, thanks, @kdvolder, it sounds fine as is.

PR #1189, opened 2022-05-31
Use the `class` `claimable` subcommands.

gmrodgers:
STILL DRAFT until someone from STK says so but making it not draft so it can be merged when I'm away on vacation.

- also includes new ClusterInstanceClass API.

Which other branches should this be merged with (if any)? N/A


Samze:
LGTM

Samze:
Rebased

PR #1192, opened 2022-06-01
Update accelerator cli docs

trisberg:
- move docs under cli-plugins
- add new fragment commands
- adjust to follow other plugin docs


snneji:
Thanks @trisberg! 

I assume this is for the 1.2 docs only, right?

trisberg:
yes, 1.2 only

PR #1193, opened 2022-06-01
Fixed image

DiegoWong:


RichardJJG:
Thanks, @DiegoWong! Just as a heads-up, you might need to fork first before raising PRs in future -- the repo permissions are changing based on roles.

PR #1203, opened 2022-06-02
Add documentation for the composition feature

ericbottard:
Fix https://github.com/pivotal/acc-release/issues/275


ericbottard:
Ping @trisberg

trisberg:
This is for TAP 1.2 docs only

PR #1208, opened 2022-06-02
Update documentation with new screenshots

svassellvmware:
Targeting documentation for 1.2 main


tap-gui/plugins/runtime-resource-visibility.md
```diff
@@ -51,23 +51,7 @@ In this page, additional information is available for Knative resources, includi
 - revisions
 - pod details
 
-   ![Screenshot of Java web app deployment page](images/runtime-resources-details.png)
```
chinigo, line 54: Did you intend to delete all this text as well?

tap-gui/plugins/runtime-resource-visibility.md
```diff
@@ -112,33 +98,29 @@ specified in `metadata.ownerReferences`. You can use this section to navigate be
 For more information about owners and dependents, see the
 [Kubernetes documentation](https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/).
 
-![Screenshot of metadata owner references](images/runtime-resources-ownership.png)
+![Screenshot of metadata owner references](images/runtime-resources-ownership-new.png)
```
chinigo, line 101: **QUESTION**

Why are all of these named as `*-new` rather than just replacing the original image?

tap-gui/plugins/runtime-resource-visibility.md
chinigo: Addressed, thank you!

PR #1215, opened 2022-06-02
Update usages of ingress domains to use new shared key introduced in https://gitlab.eng.vmware.com/tap/tap-packages/-/merge_requests/534

evankanderson:

TAP 1.2 when available.

This requires TAP 1.2.0-build.4, which should be built tonight.

### DIFF
--- a/app-live-view/install.md
+++ b/app-live-view/install.md
@@ -74,7 +74,7 @@ To install Application Live View Backend:
     ingressEnabled: "false"
     ```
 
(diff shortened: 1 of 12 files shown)


evankanderson:
Please don't merge this yet; the orginal PR got reverted due to process (not correctness) issues.

evankanderson:
This is now ready to merge -- the key is in `-build.9`.

PR #1221, opened 2022-06-06
Add RDS use case reference.

gmrodgers:
- also bump others to 0.7 (staging atm).


HenryBorys:
@gmrodgers Why are we putting links to a staging site? Before this goes to prod, you'll have to change them, which may not happen in an imperfect world. Far better to provide us with the correct link now. Can you please provide a link to a live site, or even a link that is not yet live, but will assuredly be prior to our release?

HenryBorys:
@gmrodgers We can't link to staging. @YugangW shouldn't have merged your PR, so to get your other changes in the PR, I've  remade the changes @RichardJJG made to those URLs. Presumably, you are confident the target pages will be live by our publishing date. If this is not the case, let me know, and I will put the old links back in.

RichardJJG:
Please see this thread I tagged you in last week, @HenryBorys:
https://vmware.slack.com/archives/C02P2UZHJ4F/p1654589735481809
Also, please update your version of Style Checker: it gives guidance for handling staging links.

I'll make the links staging ones again because that's what Glen would like and it makes sense to accommodate in this case.

PR #1239, opened 2022-06-08
SCST Scan - removing known CVEs from old docs according to guidelines

N00BEST:

This is intended for TAP 1.1.2

### DIFF
--- a/release-notes.md.hbs
+++ b/release-notes.md.hbs
@@ -57,6 +57,13 @@ Grype fails to find vulnerabilities during a Source Scan.
 The vulnerabilities are still found during the Image Scan,
 after the binaries are built and packaged as images.
 
+#### <a id="1-1-2-known-issues-scst-scan"></a>Supply Chain Security Tools - Scan
+
+**Error: Unable to decode cyclonedx:**
+Supply Chain Security Tools - Scan has a known issue where it will set the phase to `Error` and show an `unable to decode cyclonedx` error in the `Succeeded` condition. The root cause of the problem is not yet discovered, but it's associated with a flake that cuts the CycloneDX XML stream to the logs and then the Scan-Link controller can't process the results properly.
+
+**Workaround:** This is a flake related, if you're applying the scan manually you can delete it and re-apply it again to retry the scan. If this problem happend while running a Supply Chain from the OOTB Supply Chains, you can run `kubectl get imagescans -n <workload namespace>` or `kubectl get sourcescans -n <workload namespace>` to get the scan name, delete it running `kubectl delete <imagescan or sourcescan> <scan name> -n <workload namespace>` and the Choreographer controller will recreate it for you. 
+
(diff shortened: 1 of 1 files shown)


RichardJJG:
Hi, @N00BEST, you mention that this change is intended for TAP v1.1.2, but it deletes CVEs listed for v1.1.1.
Is this change a mistake? 
(We don't delete known issues listed for any patch -- we just document that they're fixed in a later patch/minor/major.)

N00BEST:
Hey, @RichardJJG! No, the change is not a mistake, it's taking action based on this slack thread: https://vmware.slack.com/archives/C02P2UZHJ4F/p1654704917740649?thread_ts=1654614954.425989&cid=C02P2UZHJ4F 

I also pushed some new changes for a known issue in the 1.1.2. 

Please let me know if there is any issue around this! Thank you!

RichardJJG:
Thanks, @N00BEST, that's all I needed to know before merging.

release-notes.md.hbs
```diff
@@ -57,6 +57,13 @@ Grype fails to find vulnerabilities during a Source Scan.
 The vulnerabilities are still found during the Image Scan,
 after the binaries are built and packaged as images.
 
+#### <a id="1-1-2-known-issues-scst-scan"></a>Supply Chain Security Tools - Scan
+
+**Error: Unable to decode cyclonedx:**
+Supply Chain Security Tools - Scan has a known issue where it will set the phase to `Error` and show an `unable to decode cyclonedx` error in the `Succeeded` condition. The root cause of the problem is not yet discovered, but it's associated with a flake that cuts the CycloneDX XML stream to the logs and then the Scan-Link controller can't process the results properly.
```
nedenwalker, line 63: As flake is not a technical term, let's change the sentence to something like "The root cause of the problem is not yet discovered, but it is an intermittent issue that cuts..."

release-notes.md.hbs
```diff
@@ -57,6 +57,13 @@ Grype fails to find vulnerabilities during a Source Scan.
...
+
+**Workaround:** This is a flake related, if you're applying the scan manually you can delete it and re-apply it again to retry the scan. If this problem happend while running a Supply Chain from the OOTB Supply Chains, you can run `kubectl get imagescans -n <workload namespace>` or `kubectl get sourcescans -n <workload namespace>` to get the scan name, delete it running `kubectl delete <imagescan or sourcescan> <scan name> -n <workload namespace>` and the Choreographer controller will recreate it for you.
```
nedenwalker, line 65: Similarly, "As this is an intermittent issue, if you're applying the scan manually, you can delete it...".

And there is a spelling mistake: "If this problem happened..."

release-notes.md.hbs
RichardJJG: Fixed in a separate PR

release-notes.md.hbs
RichardJJG: Fixed in a separate PR

PR #1241, opened 2022-06-09
Fix AA Conflict Resolution value

ericbottard:


RichardJJG:
Thanks, @ericbottard, does this change apply to TAP v1.1 also? (Do you want this cherry-picked to the 1-1 branch?)

ericbottard:
No sorry, this is 1.2 only

PR #1246, opened 2022-06-09
docs: add relative links and code snippet formatting

rawlingsj:


rawlingsj:
it's sometimes hard to see the markdown format changes via commit, this is the commit rendered https://github.com/pivotal/docs-tap/blob/main/tap-gui/plugins/scc-tap-gui.md

RichardJJG:
Thanks, @rawlingsj, is this change just for the TAP v1.2 docs?

rawlingsj:
np - yes just for v1.2

PR #1262, opened 2022-06-14
scc: add entry about dockerfile-based builds

cirocosta:
in tap 1.2.0, it's now possible to make use of dockerifle-based builds
by making use of a workload paramter (`--param dockerfile=...`) which
switches from the default build mechanism (kpack) to kaniko.

cc @a-peek4


snneji:
Thanks @cirocosta

PR #1263, opened 2022-06-14
Added decs about workload panel for the VScode extension

suarezjulian:
Which other branches should this be merged with (if any)? only main. feature will only be available from TAP 1.2 onwards


snneji:
Thanks @suarezjulian

PR #1269, opened 2022-06-15
Add intro to fragments to creating acc doc

trisberg:

Only for 1.2.0

### DIFF
--- a/application-accelerator/creating-accelerators/creating-accelerators.md
+++ b/application-accelerator/creating-accelerators/creating-accelerators.md
@@ -86,6 +86,12 @@ Accelerator fragments are reusable accelerator components that can provide optio
 
 The accelerator samples include three fragments - `java-version`, `tap-initialize`, and `live-update`. See the [sample-accelerators/fragments](https://github.com/sample-accelerators/fragments/tree/tap-1.2) Git repository for the content of these fragments.
 
+To discover what fragments are available to use, you can run the following command:
+
+```
+tanzu accelerator fragment list
+```
+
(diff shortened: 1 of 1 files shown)


application-accelerator/creating-accelerators/creating-accelerators.md
```diff
@@ -80,6 +80,127 @@ An alternative to using the Tanzu CLI is to create a separate manifest file and
...
+    ref:
+      branch: tap-1.2
+    url: https://github.com/sample-accelerators/fragments.git
+    subPath: java-version
+```
+
+To create the fragment (we can save the above manifest in a `java-version.yaml` file) and use:
+
+```
+tanzu accelerator apply -f ./java-version.yaml
```
gm2552, line 166: You use `tanzu acc apply` to create fragments?  This initially confused me, so if this is true, you might want to clearly call out that the above command can be used for fragments and accelerators.

gm2552, line 166: I'm seeing a mixture of kubectl and tanzu commands when applying manifest files in this document (see lines 80 and 166).  Is there a litmus test for when to use each one?

trisberg: they basically do the same thing, the accelerator apply is a bit more picky - has to be a single Accelerator or Fragment resource in the yaml

trisberg: I pushed another commit with a note to clarify

trisberg: updated to use the `accelerator apply` command consistently in this doc

PR #1270, opened 2022-06-15
Add links to Flux docs for auth and certs

trisberg:


trisberg:
This is for 1.2.0 docs

PR #1272, opened 2022-06-15
Add Fragment CRD documentation

trisberg:

This is for 1.2.0 only


snneji:
Thanks @trisberg

PR #1275, opened 2022-06-15
VSCode: Added instructions on how to see the current context and how to set the namespace for the current context

suarezjulian:
Which other branches should this be merged with (if any)? None.


RichardJJG:
Thanks, @suarezjulian!

PR #1284, opened 2022-06-17
Cartographer Conventions minor fixes

rszumlakowski:


snneji:
Thanks @rszumlakowski

PR #1289, opened 2022-06-21
Fix: updated recommended tilt version

suarezjulian:
Which other branches should this be merged with (if any)? none

Older versions had a bug where using live update would trigger a full supply chain build

Tilt issue: https://github.com/tilt-dev/tilt/pull/5694


RichardJJG:
Thanks,  @suarezjulian!

PR #1290, opened 2022-06-21
Add in label that will add workloads to gui.

gmrodgers:


RichardJJG:
Thanks, @gmrodgers, does this apply to TAP v1.1 or v1.0 as well?

gmrodgers:
Yes this can apply to TAP 1.0 and 1.1 yes! Sorry I shouldve seen the comment in the PR template about backporting!

PR #1292, opened 2022-06-22
TAP GUI customization new section

andreizimin:
Feature for 1.2.0


RichardJJG:
Thanks

PR #1294, opened 2022-06-22
Add details about shared.ingress_domain in ALV

vudayani-vmw:
It is targeted for TAP 1.2.0


RichardJJG:
Thanks, @vudayani-vmw

PR #1297, opened 2022-06-22
Update iterate-functions.md

sheltoG:
Added a step to configure the tiltfile.


RichardJJG:
Thanks, @sheltoG. This is just for the main branch, I'm guessing?

sheltoG (commented):
Actually we can take the source image part out. Let me know if I should just submit a new PR. I realized the source image part is already listed in the VScode extension instructions.

The Tiltfile has configuration fields unique to your source image and Kubernetes context. You will need to add the line `allow_k8s_contexts('cluster-name')` to your Tiltfile.

PR #1298, opened 2022-06-22
add tap gui configure tls section

dimitrybesson:
1.2 release

### DIFF
--- a/tap-gui/tls.md
+++ b/tap-gui/tls.md
@@ -31,7 +31,7 @@ In order to configure a TLS certificate for Tanzu Application Platform GUI, foll
 
         ```kubectl apply -f certificate.yaml```
 
-2. Update your `tap-gui` values to include a top level `tls` key, with subkeys for `namespace` and `secretName`. The namespace should refer to the namespace where tap-gui is running. The secret name should refer to the `secretName` value defined in your Certificate resource above.
+2. Update your `tap-gui` values to include a top level `tls` key, with subkeys for `namespace` and `secretName`. The namespace should refer to the namespace containing the above `Certificate` object. The secret name should refer to the `secretName` value defined in your Certificate resource above.
    ```yaml
     tls:
         namespace: tap-gui
--- a/tap-gui/tls.md
+++ b/tap-gui/tls.md
@@ -10,21 +10,21 @@ In order to configure a TLS certificate for Tanzu Application Platform GUI, foll
         apiVersion: cert-manager.io/v1
         kind: Issuer
         metadata:
-        name: ca-issuer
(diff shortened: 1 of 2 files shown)


dimitrybesson:
There are a lot of places throughout the docs that have hardcoded URLS starting with http://tap-gui.ingress-domain/. Now that we explicitly document the use of TLS, should these URLS be updated to reflect the possibility of them being https? Maybe something like http(s)?

tap-gui/tls.md
```diff
@@ -0,0 +1,39 @@
...
+        dnsNames:
+        - tap-gui.INGRESS-DOMAIN
+        issuerRef:
+            name: ca-issuer
+        ```
+    * add the Issuer and Certificate to your cluster:
+
+        ```kubectl apply -f certificate.yaml```
+
+2. Update your `tap-gui` values to include a top level `tls` key, with subkeys for `namespace` and `secretName`. The namespace should refer to the namespace where tap-gui is running. The secret name should refer to the `secretName` value defined in your Certificate resource above.
```
jamieklassen, line 34:
```suggestion
2. Update your `tap-gui` values to include a top level `tls` key, with subkeys for `namespace` and `secretName`. The namespace should refer to the namespace containing the above `Certificate` object. The secret name should refer to the `secretName` value defined in your Certificate resource above.
```
I may have misunderstood how the [TLSCertificateDelegation bit](https://gitlab.eng.vmware.com/project-star/pstar-backstage-poc/-/blob/develop/package-contents/config/ingress.yml) works, but I think you can put your `Certificate` in a different namespace than the one containing the tap-gui HTTPProxy and the association should still work.

tap-gui/tls.md
```diff
@@ -0,0 +1,39 @@
...
+        kind: Certificate
+        metadata:
+        name: tap-gui-cert
+        namespace: tap-gui
+        spec:
+        secretName: tap-gui-cert
+        dnsNames:
+        - tap-gui.INGRESS-DOMAIN
+        issuerRef:
+            name: ca-issuer
```
jamieklassen, line 28: Maybe indentation broke on a copy-paste?
```suggestion
        apiVersion: cert-manager.io/v1
        kind: Issuer
        metadata:
          name: ca-issuer
          namespace: tap-gui
        spec:
          selfSigned: {}
        ---
        apiVersion: cert-manager.io/v1
        kind: Certificate
        metadata:
          name: tap-gui-cert
          namespace: tap-gui
        spec:
          secretName: tap-gui-cert
          dnsNames:
          - tap-gui.INGRESS-DOMAIN
          issuerRef:
            name: ca-issuer
```

PR #1300, opened 2022-06-23
Fix indentation and variable naming.

nedenwalker:
1.2 only


RichardJJG:
Thanks, @nedenwalker

PR #1302, opened 2022-06-24
Added context for workload and deliverable

Superv0n:
Updated to indicate what the relationship is between a workload, a supply chain and a deliverable. Needs review from Allison Peek and Zach Robinson.
Still need to add graphics.


Superv0n:
no longer needed

PR #1306, opened 2022-06-24
update source image descriptions

antpensi:


RichardJJG:
Shall I cherry-pick these changes to the TAP v1.2 docs also, @antpensi?

RichardJJG:
I've had word that, yes, the relevant changes need cherry-picking, so I'll do that.

PR #1326, opened 2022-07-01
Update ingress-multicluster.md

psapna25:


psapna25:
In the new version 1.2, this is already fixed.

RichardJJG:
Thanks, @psapna25. `1-1` is, indeed, the branch that has TAP v1.1 content. We keep branch information in the [repo README](https://github.com/pivotal/docs-tap#readme).

PR #1330, opened 2022-07-05
App Live View : update cmd in doc when workload image not built with TBS

ksankaranara-vmw:


RichardJJG:
Thanks, @ksankaranara-vmw, does this apply to early TAP versions? (1.1 and/or 1.0?)

PR #1333, opened 2022-07-05
Fixed link to acc-server expose configuration

DiegoWong:

none

### DIFF
--- a/application-accelerator/vscode.md
+++ b/application-accelerator/vscode.md
@@ -1,10 +1,10 @@
 # Application Accelerator Visual Studio Code extension
-The application Accelerator Visual Studio Code extension lets you explore and generate projects 
-from the defined accelerators in Tanzu Application Platform using VSCode. 
+The application Accelerator Visual Studio Code extension lets you explore and generate projects
+from the defined accelerators in Tanzu Application Platform using VSCode.
 
 # Dependencies
-To use the vscode extension, you need to interact with the `acc-server`, for more 
-information on how to expose this server follow the instructions [here](./acc-cli.md#server-api-connections)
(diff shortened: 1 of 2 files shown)


application-accelerator/vscode.md
```diff
@@ -1,10 +1,10 @@
...
+The application Accelerator Visual Studio Code extension lets you explore and generate projects
+from the defined accelerators in Tanzu Application Platform using VSCode.
 
 # Dependencies
-To use the vscode extension, you need to interact with the `acc-server`, for more 
-information on how to expose this server follow the instructions [here](./acc-cli.md#server-api-connections)
+To use the vscode extension, you need to interact with the `acc-server`, for more
+information on how to expose this server follow the instructions [here](../cli-plugins/accelerator/overview.md#server-api-connections)
 
 # <a id="vs-code-app-accel-install"></a> Installation
```
matt-dres, line 10: Hey @DiegoWong -  A few questions/remarks
1. When you expose the port given these CLI instructions, is the user prompted with the URL to use? I see towards the end of [the linked document](https://github.com/pivotal/docs-tap/blob/main/cli-plugins/accelerator/overview.md#server-api-connections) that an example URL is provided. 
2. I noticed that there was no mention that the URL could be used for the IDE extension. I'd suggest that we add a quick note that the URL can be used for App Accelerator. Thoughts?

DiegoWong: Hey! To answer your first question, no the user is not prompted with the URL, this is a Kubernetes config, so depending where are you running the cluster, you should use one of the ways.

sounds good, I'll add a quick note in the CLI docs

DiegoWong: Added sidenote in the latest commit

PR #1335, opened 2022-07-05
mention scoped route requirement for gitlab

jamieklassen:
[TANZUSC-1414](https://jira.eng.vmware.com/browse/TANZUSC-1414)

Which other branches should this be merged with (if any)? this affects all currently released versions of TAP, so at least 1.0 and 1.1 should get this change as well.


RichardJJG:
Thanks, @jamieklassen!

PR #1342, opened 2022-07-06
fix preference path

mingxiao:


HenryBorys:
@mingxiao Is this only for 1.2 or should it also go to 1.0 or 1.1?

RichardJJG:
Thanks, @mingxiao -- I caught this one, too: 
https://github.com/pivotal/docs-tap/commit/7c5e8c7b69a23007d5774f35dfdcfa89420c1334

PR #1355, opened 2022-07-08
Added Application Single Sign-On component

irinakh:
These updates are for TAP 1.2 docs release only


RichardJJG:
Thanks, @irinakh

PR #1356, opened 2022-07-08
DRAFT: Add tap gui docs to trust nonstandard CAs

austina-lin:


RichardJJG:
Thanks, @austina-lin, does this apply to any earlier TAP v1.1. or v1.0?

PR #1358, opened 2022-07-08
Remove command that is leftover from yaml removal.

gmrodgers:
Which other branches should this be merged with (if any)? Nope


RichardJJG:
Thanks, @gmrodgers

PR #1359, opened 2022-07-08
Few small updates to STK and services

teddyking:
Hi there,

This pr contains a few small updates to consuming services section.

Thanks
Ed

Which other branches should this be merged with (if any)? N/A - target for TAP 1.2 only.

### DIFF
--- a/getting-started/about-consuming-services.md
+++ b/getting-started/about-consuming-services.md
@@ -1,7 +1,7 @@
 # About consuming services on Tanzu Application Platform
 
 As part of Tanzu Application Platform, you can work with backing services such as
-RabbitMQ, PostgreSQL, and MySQL. Binding application workloads to service instances
+RabbitMQ, PostgreSQL, and MySQL amongst others. Binding application workloads to service instances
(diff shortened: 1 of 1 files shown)


RichardJJG:
Thanks, @teddyking

teddyking:
Thanks for the notes @HenryBorys. Regarding your second point - yes essentially I wanted to convey that the list of Operators integrate well with TAP without any further modifications or configuration on the users behalf. Also noted Re: "in GitHub" usage. Thanks!

HenryBorys:
@teddyking  Is it enough to just say, "The following list of Kubernetes operators expose APIs that integrate well with Tanzu Application Platform"? Or should we say, "The following list of Kubernetes operators expose APIs that integrate well with Tanzu Application Platform without user modifications or configurations." And if you prefer the latter, consider whether it may in some cases require user modifications, in which case, we shouldn't say that. Let me know. For now we have the first option on the site.

teddyking:
I think the first option works, thanks!

PR #1369, opened 2022-07-08
Add notes about an unexpected behavior of tbs while upgrading from TA…

tylerphelan:
- When upgrading all TBS image resources will be in a failed state but the will fix themselves on the
  next build.

This is only relevant for TAP 1.2

Jira: https://jira.eng.vmware.com/browse/TANZUSC-1467

### DIFF
--- a/release-notes.md.hbs
+++ b/release-notes.md.hbs
@@ -173,6 +173,30 @@ This release has the following breaking changes, listed by area and component.
 - Breaking change 1
 - Breaking change 2
 
+#### <a id="tbs-breaking"></a> Tanzu Build Service
+
+##### <a id="tbs-breaking-upgrading"></a> Breaking Change when Upgrading TAP 1.1 to TAP 1.2
+
+**Note:** If your TAP 1.1 installation was configured with `enable_automatic_updates: false`, this breaking change can be ignored.
+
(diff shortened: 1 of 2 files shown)


tomkennedy513 (approved):

release-notes.md.hbs
```diff
@@ -173,6 +173,30 @@ This release has the following breaking changes, listed by area and component.
...
 
+#### <a id="tbs-breaking"></a> Tanzu Build Service
+
+##### <a id="tbs-breaking-upgrading"></a> Breaking Change when Upgrading TAP 1.1 to TAP 1.2
+
+**Note:** If your TAP 1.1 installation was configured with `enable_automatic_updates: false`, this breaking change can be ignored.
+
+When upgrading TAP from 1.1 to 1.2, Tanzu Build Service Image resources will automatically run a build that will fail due to a missing dependency.
+This error will not persist and any subsequent builds will resolve this error. This means that users can safely wait for the next build of their workloads which
+may be triggered by source code changes or dependency changes.
```
matthewmcnew, line 184: > or dependency changes.

Dependency changes will not trigger a build because they only rebuild READY=true images.

tylerphelan: nice

matthewmcnew (approved):

PR #1371, opened 2022-07-08
Ecr support

ryan-a-baker:


HenryBorys:
@ryan-a-baker did you talk with @piachakra about adding this to Getting started? I'm not sure how it fits or where it should go. Is there another place on the TAP TOC where this might fit?

PR #1379, opened 2022-07-12
docs: Clarification with Source Controller MavenArtifact

rashedkvm:
Signed-off-by: Rashed Kamal <krashed@vmware.com>

Which other branches should this be merged with (if any)? n/a

### DIFF
--- a/release-notes.md.hbs
+++ b/release-notes.md.hbs
@@ -60,6 +60,7 @@ This release includes the following changes, listed by component and area.
 #### <a id="src-cont-features"></a>Source Controller
 
 - Added support for pulling Artifacts from a Maven repository using the `MavenArtifact` CR.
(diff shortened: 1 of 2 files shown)


HenryBorys:
@rashedkvm Hi. Re: the promise for support in a future release: We can't make that promise. It's predicting the future, which may not happen and worst case could cause legal issues.  How important is it that they know we plan future support? Simplest is to delete it. I'm also looking for legally acceptable wording, but haven't found it yet. Let me know. Thanks.

rashedkvm:
@HenryBorys I tend to agree with your comment regarding the future release. Is there a different language to express that we are committed to deliver those future release marked items in TAP 1.3. I can create a revised PR if you can provide the content.

HenryBorys:
@rashedkvm No need for you to create another PR. I just either need to find acceptable wording, in which case, I'll replace the wording, or delete it. I've asked around and expect to have an answer tomorrow, so I'll wait till then, and then I'll let you know what I come up with. So I take it you strongly prefer to tell the customer this support is planned?

rashedkvm:
Yes, that is correct.

PR #1391, opened 2022-07-13
docs: Updating cli plugins list

warango4:
Update CLI plugins to match versions with the ones shipped with TAP v1.2


BarrAndrea (approved):
Thank you for updating the version numbers.

PR #1398, opened 2022-07-15
Customizing default route in TAP GUI

andreizimin:
functionality for TAP 1.3.0


RichardJJG:
Thanks, @andreizimin

PR #1413, opened 2022-07-25
Changes to snyk scanner installation. 

N00BEST:
Updated the referenced name for the scan-policy for consistency.

**Which other branches should this be merged with (if any)?**

This is intended for 1.2.0


RichardJJG:
Thanks

PR #1414, opened 2022-07-26
Fixes indentation in TAP GUI ca certs overlay

mstergianis:
Don't think this doc existed in 1-1. Shouldn't be relevant for TAP 1.3.x, so I think it's just 1-2.


RichardJJG:
Thanks, @mstergianis!

PR #1415, opened 2022-07-26
Add note about ALV dns to pre-requisites

kdvolder:

Anything newer than 1.2.


RichardJJG:
Thanks

PR #1417, opened 2022-07-29
Download the latest rabbitmq for GSG.

gmrodgers:
- We arent getting the latest compatibility of rmq as it bumps its
version.

Which other branches should this be merged with (if any)? 1.2.0


Samze:
LGTM

PR #1418, opened 2022-08-03
Fix missing namespace in tanzu package command

austina-lin:

Could also be merged with 1-2 and 1-1.


RichardJJG:
Thanks for the quick turnaround, @austina-lin

PR #1419, opened 2022-08-04
Showing how to get the portal logs at Learning Center

andmagom:
This change applies to all TAP versions, because I modified the troubleshooting section


RichardJJG:
Thanks, @andmagom

PR #1420, opened 2022-08-04
Update docs for pod logs in RRV tab (for TAP 1.3)

menehune23:
**Note:** This change should target the TAP 1.3 docs


RichardJJG (approved):
Thanks! I'll merge

PR #1421, opened 2022-08-05
TANZUSC-1525: allow view cluster to read mavenartifacts

garethjevans:


garethjevans:
This should also be merged with 1_2

RichardJJG:
Thanks, @garethjevans

PR #1438, opened 2022-08-17
Update docs for pod logs (TAP 1.3)

menehune23:
Adds changes for defaulting to first container, as well as stream reconnection

**Targets TAP 1.3 release**

### DIFF
--- a/tap-gui/plugins/runtime-resource-visibility.hbs.md
+++ b/tap-gui/plugins/runtime-resource-visibility.hbs.md
@@ -199,8 +199,8 @@ To see logs since a specific date and time, select or type the UTC timestamp in 
 If no logs are displayed, adjust the timestamp to an earlier time.
 If you do not select a timestamp, all logs produced since the pod was created are displayed.
 
-For optimal performance, the pod logs page limits the total log entries displayed to the most recent
-10,000 (approximately).
+For optimal performance, the pod logs page limits the total log entries displayed to the last
+10,000, at most.
 
 ### <a id="changing-log-levels"></a>Changing log levels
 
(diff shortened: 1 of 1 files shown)


tap-gui/plugins/runtime-resource-visibility.hbs.md
```diff
@@ -189,18 +189,18 @@ To resume the stream, click the **Follow Latest** button that appears after paus
...
 
-To see all logs since a specific date and time, select or type the UTC timestamp in the
+To see logs since a specific date and time, select or type the UTC timestamp in the
 **Since date** field.
 If no logs are displayed, adjust the timestamp to an earlier time.
 If you do not select a timestamp, all logs produced since the pod was created are displayed.
 
 For optimal performance, the pod logs page limits the total log entries displayed to the most recent
-10,000.
+10,000 (approximately).
```
menehune23, line 203: The reason for the "approximately" is that the 10k limit is for any log entries as well as inline notifications that appear (these notifications are generated by the UI and are not technically log "entries").

menehune23: i.e. `total_notifications + total_entries < 10,000`

tap-gui/plugins/runtime-resource-visibility.hbs.md
```diff
@@ -231,11 +231,11 @@ toggle.
 ### <a id="downloading"></a>Downloading logs
 
 To download current log content, click the **Download logs** button.
+
 For optimal performance, the pod logs page limits the total log entries downloaded to the most recent
-10,000.
+10,000 (approximately).
```
menehune23, line 236: Same comment as above for usage of the word "approximately" here. See: https://github.com/pivotal/docs-tap/pull/1438/files#r948528694

tap-gui/plugins/runtime-resource-visibility.hbs.md
menehune23: I don't think it's worth it to confuse customers with the details I mentioned, but perhaps just use the verbiage "approximate" to indicate that it may not always be exact. I welcome any thoughts/feedback though :)

menehune23: OK, I added a new commit with an alternative. Thoughts welcome.

RichardJJG:
Thanks, @menehune23

PR #1439, opened 2022-08-18
Updated ingress domain

shwethaks1:


RichardJJG:
Which other versions of TAP does this change apply to, @shwethaks1? A writer will cherry-pick accordingly.

shwethaks1:
This changes needs to updated for TAP 1.0 and 1.1 document.  I created a separate PR for both versions

PR #1441, opened 2022-08-18
Updated upgrade steps

shwethaks1:
Updated the document with more details with respect to upgrading the TAP .


RichardJJG:
@shwethaks1, which other TAP versions does this apply to, if any?

PR #1443, opened 2022-08-18
Update install-app-acc.hbs.md

x95castle1:
The default value for the server.service_type is ClusterIP and not LoadBalancer. I updated the table to match the value returned from checking the schema on the package as well (This threw me off when following the documents to find the url to configure accelerators in VSCode for example).

Running this command showed different defaults:  tanzu package available get accelerator.apps.tanzu.vmware.com/1.2.2 --values-schema --namespace tap-install

  KEY                                             DEFAULT                         TYPE     DESCRIPTION

  engine.service_type                  ClusterIP                         string   The service type for the Service of the engine.


RichardJJG:
Thanks, @x95castle1, do you know if this applies to other TAP versions? (v1.2.2, v1.2.0, etc?)

PR #1446, opened 2022-08-19
Add 1.3.0 TAP docs for App Live View - 1

ksankaranara-vmw:
This is targeted toward TAP 1.3.0 docs only.


RichardJJG:
Thanks, @ksankaranara-vmw

PR #1447, opened 2022-08-22
Document an example of how to deal with string[]

ericbottard:
Fix https://github.com/pivotal/acc-engine/issues/205

Also fix some syntax errors.

This applies to all versions of TAP where this is easily mergeable


RichardJJG:
Thanks, @ericbottard

PR #1451, opened 2022-08-23
doc: Document microservice support

mingxiao:
For the vscode extension


kiwi-bui (approved):

RichardJJG:
@mingxio, does this need cherry-picking to any earlier TAP versions, such as v1.2?

mingxiao:
@RichardJJG No it does not

PR #1459, opened 2022-08-24
Add app accelerator new features content for 1.3

trisberg:
- Using local-path when authoring
- Configuring Git-Ops managed accelerators
- Configuring Git credentials secret with custom CA certs

Which other branches should this be merged with (if any)? This is for 1.3 only.

### DIFF
--- a/application-accelerator/configuration.hbs.md
+++ b/application-accelerator/configuration.hbs.md
@@ -25,7 +25,7 @@ managed_resources:
 
 Where:
 
-- `GIT-REPO-URL` is the URL of a Git repository that contains manifest YAML files for the accelerators that you want to have managed (see below for manifest examples). You can specify a `sub_path` if necessary and also a `secret_ref` if the repository requires authentication. If not needed, then leave these additional properties out.
+- `GIT-REPO-URL` is the URL of a Git repository that contains manifest YAML files for the accelerators that you want to have managed (see below for manifest examples). You can specify a `sub_path` if necessary and also a `secret_ref` if the repository requires authentication. If not needed, then leave these additional properties out. See below for configuration of a [Git credentials secret](#creating-git-credentials).
 
 ## <a id="examples-creating-acc"></a> Examples for creating accelerators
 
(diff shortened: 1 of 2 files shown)


application-accelerator/configuration.hbs.md
```diff
@@ -7,6 +7,26 @@ Application Accelerator pulls content from accelerator source repositories using
...
+  git:
+    url: GIT-REPO-URL
+    ref: origin/main
+    sub_path: null
+    secret_ref: git-credentials
+```
+
+Where:
+
+- `GIT-REPO-URL` is the URL of a Git repository that contains manifest YAML files for the accelerators that you want to have managed (see below for manifest examples). You can specify a `sub_path` if necessary and also a `secret_ref` if the repository requires authentication. If not needed, then leave these additional properties out.
```
matt-dres, line 28: Quick suggestion - is it possible to have a hyperlink to the docs starting on line 128 so a user has an intuitive way to jump to the relevant piece of info on this?

application-accelerator/configuration.hbs.md
```diff
@@ -103,6 +123,64 @@ tanzu accelerator create my-hello-fun --git-repo https://github.com/sample-accel
...
+
+When deploying accelerators using Git repositories that need authentication and/or are installed with custom CA certificates then you need to provide some additional autentication values in a Secret. The examples in the next section provide more details about this. In this section we describe how to conventiently configure a Git credentials secret that can be used for some of the Git based examples below.
+
+You can specify the following accelerator configuration properties when installing the Application Accelerator. The same properties can be provided in the `accelerator` section of the `tap-values.yaml` file:
+
+```yaml
+git_credentials:
+  secret_name: git-credentials
+  username: GIT-USER-NAME
+  password: GIT-PASSWORD-OR-ACCESS-TOKEN
```
matt-dres, line 136: The expectation here is that an organization would have some sort of service account & its credentials would be entered here, is that right? If so, perhaps it would be beneficial to mention that "we recommend using a service account for this"?

application-accelerator/configuration.hbs.md
```diff
@@ -278,7 +356,7 @@ spec:
 
 If you are using the Tanzu CLI, then add the `--secret-ref` flag to your `tanzu accelerator create` command and provide the name of the secret for that flag.
 
-### <a id="private-suorce-image-example"></a> Examples for a private source-image repository
+### <a id="private-source-image-example"></a> Examples for a private source-image repository
```
matt-dres, line 359: nice catch! 👍

matt-dres (approved):
lgtm 👍 Few suggestions, but no blockers.

application-accelerator/configuration.hbs.md
trisberg: Added this

application-accelerator/configuration.hbs.md
trisberg: Added "We recommend using an access token for this."

PR #1462, opened 2022-08-25
Update sample policy to match Carbon Black format

asankov-cb:


YugangW:
Hi @asankov-cb, does this apply to TAP v1.2 and all patches as well? (v1.2.0, v1.2.1, v1.2.2?) Thanks.

PR #1464, opened 2022-08-25
Fix/tap gui quoted boolean

mstergianis:
Which other branches should this be merged with (if any)? Probably every active branch we have? 1.0 1.1 1.2 1.3


RichardJJG:
Thanks, @mstergianis, I've cherry-picked accordingly.

PR #1465, opened 2022-08-25
feat: Added docs to use the workload panel inside IntelliJ

suarezjulian:


RichardJJG:
Thanks, @suarezjulian, does this PR apply to any earlier versions of TAP (such as v1.2)? I'll cherry-pick if so.

suarezjulian:
@RichardJJG no, this is brand new functionality added to 1.3

PR #1474, opened 2022-08-29
docs: update instructions to use vscode context menu commands globally

kiwi-bui:


YugangW:
Hi @kiwi-bui, does this apply to TAP v1.2 and all patches as well? (v1.2.0, v1.2.1, v1.2.2?) Thanks.

YugangW:
Hi @kiwi-bui, Thanks for your confirmation. For any changes that apply to previous versions, we can do the cherry-pick for you if you could just mention it in the PR descriptions. In other words, you are not required to apply the same changes to other branches except `main` if they apply to both TAP 1.2 and TAP 1.3. Hope this answers your question. Thanks!

PR #1481, opened 2022-08-31
docs: update vs code docs for windows support

kiwi-bui:


snneji:
Hi @kiwi-bui, does this apply to any TAP versions other than 1.3?

kiwi-bui:
@snneji This should work with any TAP version, but the user will need the 0.9.0 or later of the VSCode extension.

vrabbi:
this works on linux as well. i have tested anbd been using it seamlessly on ubuntu 18, 20, and 22

kiwi-bui:
@vrabbi Thank you for the feedback, I'll let my team know!

RichardJJG:
@kiwi-bui, please submit a new PR that references Linux if you verify that it's definitely okay to tell customers that Linux is supported.

PR #1484, opened 2022-09-02
Add release notes for STK for TAP 1.0.4

teddyking:
This MR adds release notes for TAP 1.0.4, which is due to be made available on September 13, 2022. This patch release only contains one small change, which is a bump to the Services Toolkit package in order to address a CVE. As such, I simply copied the release notes for TAP 1.0.3, then modified them slightly for TAP 1.0.4 (adding a specific note about the fix for the CVE for Services Toolkit).

Ref https://gitlab.eng.vmware.com/services-control-plane/services-toolkit/-/issues/374

Which other branches should this be merged with (if any)? None.

### DIFF
--- a/release-notes.hbs.md
+++ b/release-notes.hbs.md
@@ -2,6 +2,55 @@
 
 This topic contains release notes for Tanzu Application Platform v1.0.
 
+## <a id='1-0-4'></a> v1.0.4
+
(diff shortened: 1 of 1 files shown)


RichardJJG:
Thanks for the extra info, @teddyking

PR #1491, opened 2022-09-05
Update ALV workload sample for TAP 1.2.1

ksankaranara-vmw:


RichardJJG:
@ksankaranara-vmw, should I cherry-pick this to any other docs versions?

RichardJJG:
@ksankaranara-vmw, actually, never mind -- I can see that you have very similar PRs for other branches. 
To save time in future, feel free to just PR to the most recent branch that the change applies to (usually `main`) and leave a comment saying where you want the change cherry-picked back to. It's faster for the PR author and it's easier for the writers.

PR #1493, opened 2022-09-06
Add TAP GUI release notes for TAP 1.2.2

austina-lin:


HenryBorys:
@austina-lin Just looking at the wording of your addition. I take it "bumped" means "updated," but I'm not sure what it means to "fix the severity in the Image Scanner Stage." Does this refer to a CVE? If so, we aren't referring to those in release notes. Please advise.

PR #1498, opened 2022-09-07
Adds additional API Auto Registration Documentation

jrshaffe:
Adds some additional API Auto Registration documentation that was discovered missing when looking through the preview.


karinepires (approved):
Thank you Joel for fixing the missing spots on the documentation for API Auto Registration 🎉 .

PR #1502, opened 2022-09-07
vscode: debug command from command palette

kiwi-bui:


RichardJJG:
Thanks, @kiwi-bui, which other versions of the TAP docs does this change apply to, if any?

kiwi-bui:
@RichardJJG Hi, this is only for 1.3. Are there separate processes for doc updates for all versions vs just the newest version?

RichardJJG:
Thanks, and the protocol is for the PR author to make changes to only the latest docs version possible. If a change applies to all versions of TAP, for example, you PR to the `main` branch and then you tell us in a comment here how far back you want a writer to cherry-pick. (See the very first comment on this page.) It saves everyone time that way.
A branch and docs version table is on the [README](https://github.com/pivotal/docs-tap#readme).

PR #1505, opened 2022-09-07
Reduce RabbitMQ focus in getting-started services docs

ryankilroy:
Modifies the walkthrough example to include references to the equivalent setup items for Postgres and MySQL for Tanzu

_Notes:_ 
- `getting-started/consume-services.hbs.md` still needs updating to be consistent with this change


getting-started/set-up-services.hbs.md
```diff
@@ -37,22 +37,25 @@ Before following this walkthrough, you must:
 
 1. Have access to a cluster with Tanzu Application Platform installed.
 1. Have downloaded and installed the Tanzu CLI and the corresponding plug-ins.
-1. Ensure your Tanzu Application Platform cluster can pull the images required by the RabbitMQ Cluster Kubernetes operator. For more information, see the [RabbitMQ documentation](https://www.rabbitmq.com/kubernetes/operator/using-operator.html).
+1. Ensure your Tanzu Application Platform cluster can pull the images required by the service operator. For more information, see the following documentation:
```
st3v, line 40: *nitpick*: Might be good to be more specific here w/r to the term _"images"_. I know this is nothing that has ben introduced by this PR and was there before, still it caught my eye when reading over this.

```suggestion
1. Ensure your Tanzu Application Platform cluster can pull the container images required by the service operator. For more information, see the following documentation:
```

getting-started/set-up-services.hbs.md
```diff
@@ -37,22 +37,25 @@ Before following this walkthrough, you must:
...
+   * [RabbitMQ Cluster Kubernetes operator](https://www.rabbitmq.com/kubernetes/operator/using-operator.html).
+   * [VMware Tanzu SQL with Postgres for Kubernetes](https://docs.vmware.com/en/VMware-Tanzu-SQL-with-Postgres-for-Kubernetes/index.html).
```
st3v, line 42: *nitpick*: It's not clear to me why this doc is all of the sudden sending me to the Tanzu Postgres and MySQL docs even though it told me earlier that it is using RabbitMQ as the example. I suppose this is to highlight the fact that there are other Kubernetes operators and the use different container images? If so, maybe this could be stated more clearly? 

Orthogonally, the reader might wonder why we picked these two services specifically, in addition to RabbitMQ. Is it because they are "Tanzu services", i.e. part of the family? Or did we pick them because those are compatible with the binding spec? Again, seems like we could add a bit more context here.

getting-started/set-up-services.hbs.md
```diff
@@ -37,22 +37,25 @@ Before following this walkthrough, you must:
...
+> **Note:** Although this walkthrough uses the RabbitMQ Cluster Kubernetes operator
+> as an example, the set up steps remain mostly the same for any compatible operator.
+
 This section covers the following:
 
-* Installing the RabbitMQ Cluster Kubernetes operator.
+* Installing the selected service Kubernetes operator.
 * Creating the role-based access control (RBAC) rules to grant Tanzu Application Platform permission to interact
-with the newly-installed APIs provided by the RabbitMQ Cluster Kubernetes operator.
+with the newly-installed APIs provided by the operator.
```
st3v, line 54: **nitpick** adding more context to the overloaded term _"operator"_
```suggestion
with the APIs provided by the newly-installed Kubernetes operator.
```

getting-started/set-up-services.hbs.md
st3v: Seeing Postgres and MySQL mentioned again later in this doc. Are we really still only using RabbitMQ as an example here? Or are we in fact using three examples, i.e. RMQ, Postgres, and MySQL? Maybe it's just me but seeing Postgres and MySQL in all of the remaining steps of this doc doesn't seem to align with the preamble giving earlier. Maybe we should be more clear up there? Potentially some thing along the lines of...

* this doc shows how binding compatible services can be consumed from TAP workloads
* it uses Tanzu RabbitMQ, Postgres, and MySQL as examples
* the same process applies to any other service that is compatible with the binding spec
* for services that are not compatible with the binding spec, check out these examples (link to cloud-managed service use cases or something else that makes use of `SecretTemplate`, e.g. Greg's blog post)

ryankilroy: The intention was to provide walkthrough value to folks who were going through the Readme for other services, but get turned off it by the fact that it didn't address specifics for their service. Though I think we might need to make the decision of whether or not we want to walk through more than one example or not. I agree with you that there is now a discrepancy between the preamble and the examples, but I'm unsure how to address the issue properly short of providing an example walkthrough for every service we support. I like the idea of adding the reference to the `SecretTemplate` example

ryankilroy: After discussion, we agreed this isn't really a long term solution. A rewrite of the docs to more clearly cover all our use-cases may be required, but this hopefully is an improvement for now.

HenryBorys:
@ryankilroy Is this ready to merge?

HenryBorys:
@ryankilroy Who do you want to review this?

ryankilroy:
@HenryBorys it should be good to merge! @st3v reviewed it with me offline

HenryBorys:
@ryankilroy Also, is this for 1.2 and 1.3 or just 1.3?

PR #1509, opened 2022-09-07
vscode: update workload apply command instructions

kiwi-bui:


RichardJJG:
Thanks, @kiwi-bui, which other versions of the TAP docs does this change apply to, if any?

kiwi-bui:
@RichardJJG No other versions. Thank you!

PR #1521, opened 2022-09-09
AppSSO: add docs for upgrading to 2.0

sarahwalther:


RichardJJG:
Thanks, @sarahwalther, does this content apply to just the TAP v1.3 docs or should I cherry-pick it back to earlier versions?

PR #1523, opened 2022-09-12
Update TAP profile config for ALV and version

ksankaranara-vmw:


ksankaranara-vmw:
This is for 1.3.0 Release only.

RichardJJG:
Thanks, @ksankaranara-vmw

PR #1526, opened 2022-09-12
VSEE-698 | Generalize scanner installation

chaomonica:
- main


RichardJJG:
Thanks, @chaomonica

PR #1530, opened 2022-09-12
appsso: fix language explaining openshift SCCs

sarahwalther:


RichardJJG:
Thanks, @sarahwalther

PR #1533, opened 2022-09-13
Remove duplicate steps for CBC scanner install

asankov-cb:

Just `main`


RichardJJG:
Thanks, @asankov-cb

PR #1534, opened 2022-09-13
add tap-auth release notes for 1.3.x

dherbrich:
Which other branches should this be merged with (if any)? the release notes should be added to tap 1.3.0 docs.


RichardJJG:
Thanks, @dherbrich

PR #1536, opened 2022-09-13
appsso: fix accelerator location in workload creation (tutorial)

ddubson:


RichardJJG:
Thanks, @ddubson, does this change apply to any earlier TAP docs versions?

PR #1538, opened 2022-09-13
Added known issue to vscode and intellij plugins

suarezjulian:
Which other branches should this be merged with (if any)? Also applies to 1.2


suarezjulian:
@corythorpe fyi

RichardJJG:
Thanks, @suarezjulian

PR #1539, opened 2022-09-13
Added iterate cluster as one of the profiles that come with api auto …

jrshaffe:
Adds the "iterate" cluster to the profiles mentioned that get api auto registration by default

Which other branches should this be merged with (if any)? TAP 1.3


RichardJJG:
Thanks, @jrshaffe

PR #1540, opened 2022-09-13
SCST Policy Controller known issue

DennyHoang:

Also cherry-pick to main for TAP 1.3.0 so that it is noted that TAP 1.2.2 and earlier contains this known issue.

We plan on making a new release for the known issue for TAP 1.3.0 which should resolve it.


DennyHoang:
Currently, I added Known Issue to the release notes and the overview of SCST Policy Controller. Are there any other places that I should consider adding known issues to?

RichardJJG:
Thanks, @DennyHoang, it's actually best to just add the known issue to the release notes. Where some workaround is needed, we ask that the engineer writes the steps in the appropriate topic (usually Troubleshooting) and that the engineer links to those steps from the known issue: "For more information, see \[TOPIC-TITLE\]\(URL\)."

HenryBorys:
@DennyHoang Re: cherry picking this to 1.3: If the issue is resolved in 1.3, it's no longer a known issue for 1.3, so doesn't quite make sense to include it as a known issue in 1.3.  It will be noted in the Resolved issues section for 1.3 that this known issue for 1.2.2 and earlier is now resolved, so it will be referenced there. If this known issue is NOT resolved in 1.3, then we can include it in the 1.3 known issues. Please let me know if it is not resolved as 1.3 approaches.
On a separate note,  I do understand the problem users have with how our known issues are presented, which requires them to review all past release notes to understand the history of known issues and in which release they were resolved. Very inconvenient. I'm working on a separate solution to this. Hopefully to be implemented in 1.3 release notes.

PR #1543, opened 2022-09-13
add warning for default roles

dherbrich:


RichardJJG:
Thanks, @dherbrich, should a writer cherry-pick this to any other TAP docs versions?

dherbrich:
Hi. this applies to any TAP version containing tap-auth.  so it could be merged back to the point where this docs page was created.

PR #1546, opened 2022-09-14
Update SCG and Steeltoe docs in ALV for TAP 1.3

ksankaranara-vmw:
Which other branches should this be merged with (if any)


ksankaranara-vmw:
> Thanks, @ksankaranara-vmw, is this just for TAP v1.3?

yes only TAP 1.3

PR #1550, opened 2022-09-14
Added part-of labels to scan section to ensure workload loads in GUI.

nedenwalker:
1.3.x (if possible), otherwise just main


RichardJJG:
Thanks, @nedenwalker

PR #1551, opened 2022-09-14
Improve Azure AD docs

elfotografo007:
This change also applies to 1.2.2.


RichardJJG:
Thanks, @elfotografo007

PR #1553, opened 2022-09-14
VSEE-740 More precisely specified params and examples.

nedenwalker:
1.3.x


RichardJJG:
Thanks, @nedenwalker

PR #1556, opened 2022-09-15
Add docs for service-operator role

teddyking:
This PR adds information about the new "Service Operator" user role, which is available as of TAP 1.3. Note that I have not included updates to the [Detailed role permissions breakdown](https://github.com/pivotal/docs-tap/blob/main/authn-authz/permissions-breakdown.hbs.md) page as I believe this page is due to be either reworked or removed in the near future (cc @dherbrich).

Which other branches should this be merged with (if any)? None


RichardJJG:
Thanks, @teddyking

PR #1557, opened 2022-09-15
Small updates to consuming services docs

teddyking:
A few small updates to the consuming services guide (mostly just updating links, as well as tidy up of some of the commands).

Which other branches should this be merged with (if any)? 1.2

### DIFF
--- a/getting-started/consume-services.hbs.md
+++ b/getting-started/consume-services.hbs.md
@@ -73,8 +73,6 @@ To create an application workload:
     Expected output:
 
     ```console
-      Warning: This is an ALPHA command and may change without notice.
-
       NAME   READY  REASON
       rmq-1  True
     ```
@@ -88,8 +86,6 @@ To create an application workload:
     Expected output:
 
     ```console
-      Warning: This is an ALPHA command and may change without notice.
-
     Name: rmq-1
     Status:
       Ready: True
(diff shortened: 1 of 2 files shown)


RichardJJG:
Thanks, @teddyking

HenryBorys:
@teddyking Those links are not yet working, so they can't be published to 1.2 until the target pages are live. Do you know when those will go live? In the meantime, I have to revert the links back to the earlier version on 1.2. Please let me know when you expect the links to go live so I can be sure they are live before we publish 1.3. Thanks.

HenryBorys:
@teddyking Going forward, let's not update links in previous tap versions to current targets at all, unless the past targets are actually broken, and then hopefully the new targets are live. Past versions are a snapshot in history so to speak, and we only need to fix things that are actually broken for that release, not update to the latest and greatest just because we now can. Our next release can give the latest and greatest. If it's critical we update something, that's another story. Thanks.

RichardJJG:
@teddyking, @HenryBorys,

@teddyking is right to xref the v0.8 docs here because this is the main branch: TAP and Services Toolkit both release on the same day. So, when these docs go live so, too, will the v0.8 docs. We just have to turn the links @teddyking has provided into staging links so people doing QA on the docs can see the xrefs in the meantime.

As part of the release, Yugang will convert staging links such as these into prod links, as described on our 
[Release Day Confluence page](https://confluence.eng.vmware.com/display/MIX/New+TAP+Release+Publishing+Tasks#NewTAPReleasePublishingTasks-Docs).

When @teddyking says "merge with the v1.2 docs", he's talking about the PR as a whole and not necessarily the xref links. He trusts the writer to know what not to merge in a large PR such as this (and to ask an engineer when unsure). 

We DO change xrefs in older docs to the latest and greatest versions if that's what's in the customer's interests (but it rarely is because of compatibility). 

@teddyking, when we merge this PR with the v1.2 docs, do you want us to change the v0.8 Services Toolkit docs xref links to the v0.7 links (in keeping with the other Services Toolkit xrefs in TAP v1.2)?

teddyking:
(sorry for the late reply here I've been OOO) - thanks @RichardJJG your summary above accurately covers my intentions for this PR. I would like both the TAP 1.2 and TAP 1.3 docs to point to the updated v0.8 xref links (there is no breaking change there, only incremental improvements that will benefit readers of the TAP 1.2 docs as well), which I guess means we'll have to wait until those docs go live before making the change on the TAP 1.2 branch?

RichardJJG:
@teddyking, yes, we'll have to wait for the v0.8 docs to go live before cross-referencing them from the TAP v1.2 docs. 
I've created a [story](https://www.pivotaltracker.com/story/show/183403241) for this and blocked it until October 13 (the release date for TAP v1.3 docs and Services Toolkit v0.8 docs).

CC @HenryBorys

PR #1558, opened 2022-09-15
chore [META-769] : Updated metadata store api documents

shaileshpandey:
Which other branches should this be merged with (if any)? No. Changes are for TAP 1.3


RichardJJG:
Thanks, @shaileshpandey

PR #1560, opened 2022-09-15
Added legacy SBOM deprecation notice.

ncarlson:
https://github.com/pivotal/build-service/issues/870

### DIFF
--- a/tanzu-build-service/install-airgapped.hbs.md
+++ b/tanzu-build-service/install-airgapped.hbs.md
@@ -25,6 +25,14 @@ Approximately 10&nbsp;GB of registry space is required when using the `full` dep
 
 - Your Docker registry must be accessible with user name and password credentials.
 
+## <a id='DeprecatedFeatures'></a> Deprecated Features
+
+* Users are encouraged to use the SBOM format published in
+  [Buildpacks API 0.7.0](https://buildpacks.io/docs/reference/spec/migration/buildpack-api-0.6-0.7/#new-standardized-sbom-format).
+  Older (legacy) SBOM formats are deprecated and are disabled by default in TBS. Support for legacy SBOM formats be
+  enabled by either adding `include_legacy_bom: true` to the `tbs-values.yml` file or by passing
+  `--data-value-yaml include_legacy_bom=true` to `ytt` when processing the TBS bundle.
+
 ## <a id='tbs-offline-install-package'></a> Install the Tanzu Build Service package
 
 These steps assume that you have installed the Tanzu Application Platform packages
--- a/tanzu-build-service/install-tbs.hbs.md
+++ b/tanzu-build-service/install-tbs.hbs.md
@@ -24,6 +24,14 @@ Approximately 10&nbsp;GB of registry space is required when using the `full` dep
 
 - Your Docker registry must be accessible with user name and password credentials.
 
+## <a id='DeprecatedFeatures'></a> Deprecated Features
+
+* Users are encouraged to use the SBOM format published in
+  [Buildpacks API 0.7.0](https://buildpacks.io/docs/reference/spec/migration/buildpack-api-0.6-0.7/#new-standardized-sbom-format).
+  Older (legacy) SBOM formats are deprecated and are disabled by default in TBS. Support for legacy SBOM formats be
+  enabled by either adding `include_legacy_bom: true` to the `tbs-values.yml` file or by passing
+  `--data-value-yaml include_legacy_bom=true` to `ytt` when processing the TBS bundle.
+
 ## <a id='tbs-tcli-install'></a> Install the Tanzu Build Service package
 
 To install Tanzu Build Service by using the Tanzu CLI:


tanzu-build-service/install-airgapped.hbs.md
```diff
@@ -25,6 +25,14 @@ Approximately 10&nbsp;GB of registry space is required when using the `full` dep
 
 - Your Docker registry must be accessible with user name and password credentials.
 
+## <a id='DeprecatedFeatures'></a> Deprecated Features
+
+* Users are encouraged to use the SBOM format published in
+  [Buildpacks API 0.7.0](https://buildpacks.io/docs/reference/spec/migration/buildpack-api-0.6-0.7/#new-standardized-sbom-format).
+  Older (legacy) SBOM formats are deprecated and are disabled by default in TBS. Support for legacy SBOM formats be
+  enabled by either adding `include_legacy_bom: true` to the `tbs-values.yml` file or by passing
```
tomkennedy513, line 33: They are still enabled by default so the user would pass false to disable them

tylerphelan:
let's add the notice here as well: https://github.com/pivotal/docs-tap/blob/main/release-notes.hbs.md

let's also mention that it will be removed by default in tap 1.4 and removed entirely from tap 1.5

tanzu-build-service/install-airgapped.hbs.md
```diff
@@ -25,6 +25,14 @@ Approximately 10&nbsp;GB of registry space is required when using the `full` dep
 
 - Your Docker registry must be accessible with user name and password credentials.
 
+## <a id='DeprecatedFeatures'></a> Deprecated Features
+
+* Users are encouraged to use the SBOM format published in
+  [Buildpacks API 0.7.0](https://buildpacks.io/docs/reference/spec/migration/buildpack-api-0.6-0.7/#new-standardized-sbom-format).
+  Older (legacy) SBOM formats are deprecated and are disabled by default in TBS. Support for legacy SBOM formats be
```
tylerphelan, line 32: same comment as here: https://github.com/pivotal-cf/docs-build-service/pull/307#discussion_r972295910

tomkennedy513:
> let's add the notice here as well: https://github.com/pivotal/docs-tap/blob/main/release-notes.hbs.md
> 
> let's also mention that it will be removed by default in tap 1.4 and removed entirely from tap 1.5

it actually we may have to keep it on by default in 1.4, off by default in 1.5 and removed in 1.6 as per https://confluence.eng.vmware.com/display/CNA/TAP+Deprecation+Policy#TAPDeprecationPolicy-GenerallyAvailable(stable) if we consider 1.3 to be our intent to deprecate or not

tanzu-build-service/install-airgapped.hbs.md
ncarlson: Thanks, @tomkennedy513 . I reverted this PR and opened a new draft PR here: https://github.com/pivotal/docs-tap/pull/1570.

ncarlson:
> let's add the notice here as well: https://github.com/pivotal/docs-tap/blob/main/release-notes.hbs.md
> 
> let's also mention that it will be removed by default in tap 1.4 and removed entirely from tap 1.5

Good idea! I reverted this PR and opened a new draft PR here: https://github.com/pivotal/docs-tap/pull/1570.

tanzu-build-service/install-airgapped.hbs.md
ncarlson: Thanks @tylerphelan, I reverted this PR and opened a new draft PR here with this change: https://github.com/pivotal/docs-tap/pull/1570.

ncarlson:
> > let's add the notice here as well: https://github.com/pivotal/docs-tap/blob/main/release-notes.hbs.md
> > let's also mention that it will be removed by default in tap 1.4 and removed entirely from tap 1.5
> 
> it actually we may have to keep it on by default in 1.4, off by default in 1.5 and removed in 1.6 as per https://confluence.eng.vmware.com/display/CNA/TAP+Deprecation+Policy#TAPDeprecationPolicy-GenerallyAvailable(stable) if we consider 1.3 to be our intent to deprecate or not. we should get clarity on that though

Thanks @tomkennedy513. If we read the text verbatim, it would appear that we can't actually add the `include_legacy_bom` flag in TAP 1.3, as we would have had to announce its deprecation in TAP 1.2. But if we're deprecating and including in the same feature (accelerated deprecation?), I would think that the 1.4->disable, 1.5->remove plan would be in accordance. Any thoughts, @tylerphelan or @matthewmcnew? Lets continue this discussion in the new PR: https://github.com/pivotal/docs-tap/pull/1570

From the policy:

### <VERSION> (legacy BOM deprecation announced in 1.2 or 1.3?)
* Before or concurrent with this version, **provide internal and external announcements of the intent to deprecate the feature.**

### <VERSION> + 1 / 3 months (`include_legacy_bom: true` added in 1.3 or 1.4?)
* If a succeeding version or alternative feature is being provided, **it should be added**. (I.e., new SBOM support added)
* If a succeeding version or alternative feature is being provided, **the documentation and release notes should describe it**.

### <VERSION> + 2 / 6 months (`include_legacy_bom: false` added in 1.4 or 1.5?)
* If a succeeding version or alternative feature is being provided, **it should be configured as the default way** to achieve the user’s desired outcome. (I.e., `include_legacy_bom: false`)
* If a succeeding version or alternative feature is being provided, the documentation and release notes **should indicate that it has been configured as the default way to achieve the user’s desired outcome**. 

### <VERSION> + 3 / 9 months (`include_legacy_bom` and legacy BOM support removed in 1.5 or 1.6?)
* **Feature is removed.**
* Documentation and release notes should state that the feature has been removed.

PR #1563, opened 2022-09-15
Revert "Added legacy SBOM deprecation notice."

ncarlson:
Reverts pivotal/docs-tap#1560


RichardJJG:
Thanks, @ncarlson, does this change apply only to the TAP v1.3 docs?

snneji:
Hi, is the reason for reverting this because you now plan to deprecate in 1.4 instead of 1.3?

ncarlson:
> Hi, is the reason for reverting this because you now plan to deprecate in 1.4 instead of 1.3?

No, I'm reverting because the merge happened prematurely, which is my fault. The PR should have been a draft PR. I need to respond to comments/change requests before merging.

> Thanks, @ncarlson, does this change apply only to the TAP v1.3 docs?

Yes.

PR #1566, opened 2022-09-16
Update metadata-store service account secret generation docs

julia-pu:
TAP 1.3


RichardJJG:
Thanks, @julia-pu

PR #1570, opened 2022-09-16
Deprecation of legacy bom

ncarlson:

None.

Addresses issues raised in https://github.com/pivotal/docs-tap/pull/1560.

### DIFF
--- a/release-notes.hbs.md
+++ b/release-notes.hbs.md
@@ -47,7 +47,7 @@ This release includes the following changes, listed by component and area.
 
 - [Supply Chain Security Tools - Sign](scst-sign/overview.md) is deprecated. For migration information, see [Migration From Supply Chain Security Tools - Sign](./scst-policy/migration.hbs.md).
 
-- The (legacy) CNB BOM format is deprecated, but enabled by default in TBS. In TAP 1.4, support will be disabled by 
+- The (legacy) CNB BOM format is deprecated, but is enabled by default in TBS. In TAP 1.4, support will be disabled by 
 default. And in TAP 1.5, support will be removed. To manually disabled legacy CNB BOM support add 
 `include_legacy_bom=false` to the `tbs-values.yml` file or to the `tap-values.yml` file under the `buildservice:`
 stanza.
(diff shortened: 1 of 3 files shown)


ncarlson:
From the deprecation policy:

> ### `<VERSION>` (legacy BOM deprecation announced in 1.2 or 1.3?)
> * Before or concurrent with this version, **provide internal and external announcements of the intent to deprecate the feature.**
> 
> ### `<VERSION>` + 1 / 3 months (`include_legacy_bom: true` added in 1.3 or 1.4?)
> * It is expected that **the feature will continue to operate without change** during the lifetime of this version.
> * If a succeeding version or alternative feature is being provided, **it should be added**. 
> * If a succeeding version or alternative feature is being provided, **the documentation and release notes should describe it**.
> 
> ### `<VERSION>` + 2 / 6 months (`include_legacy_bom: false` added in 1.4 or 1.5?)
> * If a succeeding version or alternative feature is being provided, **it should be configured as the default way** to achieve the user’s desired outcome.
> * If a succeeding version or alternative feature is being provided, the documentation and release notes **should indicate that it has been configured as the default way to achieve the user’s desired outcome**. 
> 
> ### `<VERSION>` + 3 / 9 months (`include_legacy_bom` and legacy BOM support removed in 1.5 or 1.6?)
> * **Feature is removed.**

From the text above, my understanding is:
* "Feature" = "legacy BOM format support"
* "alternative feature" = "SBOM format support"
* <Version> is 1.3, but should have been 1.2.
* The "_before or current with this version_" makes this a bit ambiguous. If we're announcing deprecation of legacy BOM format support in 1.3, then:
  * In 1.4 (1.3+1), and only in 1.4 should _SBOM_ format support be introduced, but disabled by default. 
  * In 1.5 (1.3+2), SBOM format support should be enabled by default.
  * In 1.6 (1.3+3), legacy BOM support is removed entirely
* If we're saying we'll announce deprecation of BOM support _and_ introduce SBOM support in 1.3, then enabled by default should occur in 1.4, and removal should happen in 1.5

Questions:
* Does the above sound right?
* To be in accordance with "If a succeeding version or alternative feature (_I.e., SBOM_) is being provided, the documentation or release notes should describe it", should we link to https://buildpacks.io/docs/tools/pack/cli/pack_sbom_download/ in the description?

tylerphelan (approved):

matthewmcnew:
+ 2 / 6 months
> If a succeeding version or alternative feature is being provided, the documentation and release notes should indicate that it has been configured as the default way to achieve the user’s desired outcome

I think therefore it needs to remain the default until 1.5.

snneji:
Replaced by #1705

PR #1571, opened 2022-09-16
Update TAP 1.3 documentation for `server` and `worker` workloads

evankanderson:

TAP 1.3 if already cut.


workloads/server.hbs.md
```diff
@@ -0,0 +1,98 @@
...
+
+* Store state locally
+* Run background tasks outside of requests
+* Provide multiple network ports or non-HTTP protocols
+* Are not a good match for the `web` workload type
+
+Applications using the `tcp` workload type have the following features:
+
+* Do not natively autoscale, but can be used with the Kubernetes Horizontal Pod Autoscaler
+* By default are exposed only within the cluster using a `ClusterIP` Service
```
tshaine, line 23: Is there documentation we can link that shows the user how to configure a `ClusterIP` Service?

workloads/server.hbs.md
```diff
@@ -0,0 +1,98 @@
...
+  ...
+```
+
+The `ports` parameter can be set from the `tanzu apps workload create` command line as `--param-yaml 'ports=[{"port": 8080}]'`.
+
+The following values are valid within the `ports` argument:
+
+| field | value |
+|-------|-------|
+| `port` | The port on which the application should be exposed to the rest of the cluster |
```
tshaine, line 96: Is it required to specify a port in a server workload? I assume it isn't required but calling it out just in case

workloads/server.hbs.md
evankanderson: The workload will configure the `ClusterIP` service itself; the user doesn't need to do anything but use the workload type.

workloads/server.hbs.md
evankanderson: See line 66 -- conventions can fill these in of bit specified; any thoughts on how to represent this?

(There's a little mix of how-to and reference docs here}

tshaine: Ah, that didnt click the first time I read that. That makes sense though. Can the user get into a situation where their application isn't secure by default because they failed to define a port in the container conventions for their cluster? If so, we may want to add an example suggested config and/or a kubectl command that will show what ports are exposed.

evankanderson: The port is exposed only within the cluster, and only for `server` type workloads.

The next defense would be to define NetworkPolicy to restrict within-cluster TCP access (because `server` workloads might or might not be HTTP).

workloads/server.hbs.md
```diff
@@ -0,0 +1,98 @@
+# Using Server workloads
+
+## <a id="overview"></a>Overview
+
+The `tcp` workload type allows you to deploy traditional network applications on
```
zrob, line 5: should tcp be server now ?

workloads/server.hbs.md
```diff
@@ -0,0 +1,98 @@
...
+Tanzu Application Platform.
+Using an application workload specification, you can build and deploy application
+source code to a manually-scaled Kubernetes deployment which exposes an in-cluster Service endpoint.
+If required, you can use environment-specific LoadBalancer Services or Ingress resources to
+expose these applications outside the cluster.
+
+The `tcp` workload is a good match for traditional applications, including HTTP applications,
```
zrob, line 12: tcp->server?

workloads/server.hbs.md
```diff
@@ -0,0 +1,98 @@
...
+that are implemented as follows:
+
+* Store state locally
+* Run background tasks outside of requests
+* Provide multiple network ports or non-HTTP protocols
+* Are not a good match for the `web` workload type
+
+Applications using the `tcp` workload type have the following features:
```
zrob, line 20: tcp->server?

PR #1575, opened 2022-09-19
Clearer language on PR flow requirements

waciumawanjohi:
This also applies to 1.2


RichardJJG:
Thank you

PR #1576, opened 2022-09-19
appsso: require tokenSignature

sarahwalther:
This change should _not_ go into the `1.3` release. I'm not quite sure which branch I should be opening this pull request against.


RichardJJG:
@sarahwalther, should it go in the TAP v1.4 docs? The v1.3.1 docs?

PR #1577, opened 2022-09-19
contour install docs: sync sa name in example pkgis

KauzClay:
* while installing contour with this doc, I was running into problems with the pkgi not reconciling. Turns out, the two example pkgis for contour don't have the same SA name. I copied the first one, and didn't realize that the sa name was different.

main

### DIFF
--- a/cert-mgr-contour-fcd/install-cert-mgr.hbs.md
+++ b/cert-mgr-contour-fcd/install-cert-mgr.hbs.md
@@ -209,7 +209,7 @@ To install Contour from the Tanzu Application Platform package repository:
       name: contour
       namespace: tap-install
     spec:
-      serviceAccountName: tap-install-sa
+      serviceAccountName: contour-tap-install-sa
       packageRef:
         refName: contour.tanzu.vmware.com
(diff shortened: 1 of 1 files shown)


RichardJJG:
Thanks, @KauzClay, and you're saying this definitely applies to just `main`? I ask because it sounds like it might apply to production version(s) of TAP.

KauzClay:
ah @RichardJJG my bad, it does look like it applies to 1.2.2. and possibly older versions

What is the process for updating older versions? Do they only update when a new version (patch or minor) of TAP happens?

I was under the impression that TAP 1.2.2 was already out, so I thought I should just fix it in main so it is in TAP 1.3.0

RichardJJG:
> What is the process for updating older versions?
You mention in a comment on the PR which older versions your changes also apply to. A writer then cherry-picks the changes to the relevant branches after editing the original commit.

RichardJJG:
@KauzClay, we'll cherry-pick to the TAP v1.2 docs. Please let us know if it applies to TAP v1.1 and v1.0 also.

PR #1579, opened 2022-09-19
Update ootb docs relating to scan

rxl7906:
Summary:
* Updating docs pertaining to scan/policy

N/A


RichardJJG:
Thanks

PR #1580, opened 2022-09-19
Update tap docs and release notes for new contour default

KauzClay:
* a default parameter in TAP for Contour has changed in TAP 1.3.0, which can cause some issues when installing on Kind

### DIFF
--- a/cert-mgr-contour-fcd/install-cert-mgr.hbs.md
+++ b/cert-mgr-contour-fcd/install-cert-mgr.hbs.md
@@ -301,9 +301,15 @@ To install Contour from the Tanzu Application Platform package repository:
             infrastructure_provider: aws
         ```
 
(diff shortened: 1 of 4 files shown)


KauzClay:
I'm not 100% sure that these changes are in the right place. I'm trying to inquire more about it here: https://vmware.slack.com/archives/C02P2UZHJ4F/p1663625516540819

RichardJJG:
> I'm not 100% sure that these changes are in the right place. I'm trying to inquire more about it here: https://vmware.slack.com/archives/C02P2UZHJ4F/p1663625516540819

Thanks, @KauzClay, please comment here when it's ready to merge

KauzClay:
@RichardJJG based on @YugangW 's comments on slack, sounds like this is ready to go

PR #1581, opened 2022-09-20
Update App Live View docs

ksankaranara-vmw:
Changes targeted only for TAP 1.3 release.


RichardJJG:
Thanks

PR #1583, opened 2022-09-20
Add Telemetry docs to TAP 1.3

xtreme-shane-lattanzio:
Which other branches should this be merged with (if any)? These changes are only meant for 1.3 and beyond


RichardJJG:
Thanks, @xtreme-shane-lattanzio

PR #1586, opened 2022-09-21
Updates some Buildpack references to match current state

ryanmoran:
None.


snneji:
@mgibson1121 I can take a look. Which link/table would you like to provide?

Sorry I missed this, I was out some of that week and am only getting to my messages.

PR #1589, opened 2022-09-22
appsso: fix status label

sarahwalther:


RichardJJG:
@sarahwalther, should a writer cherry-pick this to any live versions of the TAP docs, such as v1.2?

PR #1590, opened 2022-09-22
Add extended details for API Auto Registration for better dev experience

shrutiyer:
This PR is to be merged for TAP 1.3 docs


shrutiyer:
@csterwa can you check and ensure it represents our package well? Happy to add/edit any thing to make it more user friendly

Any suggestions @LittleBaiBai @jrshaffe @visomar @tiagofernandez?

api-auto-registration/installation.hbs.md
```diff
@@ -2,20 +2,18 @@
 
 This document describes how to install API Auto Registration from the Tanzu Application Platform package repository.
 
->**Note:** Use the instructions on this page if you do not want to use the iterate, run, or full profile to install packages.
-The "iterate", "run", and "full" profiles include API Auto Registration by default.
+>**Note:** The "iterate", "run", and "full" profiles include API Auto Registration by default. 
+> If your cluster is one of these profiles, skip the install and proceed to the [Usage section](usage.md).
```
LittleBaiBai, line 6: For `run` profile clusters, their installation is probably not going to work unless they specified `api_auto_registration.tap_gui_url` and `cluster_name` in their `tap-values.yaml`, so I'm worried about us telling users to skip the section that talks about values altogether. However, I'm not sure what's the best way to point that out.

api-auto-registration/installation.hbs.md
```diff
@@ -79,10 +71,12 @@ To install API Auto Registration:
     kubectl get secret tap-values -n tap-install -o jsonpath="{.data['tap-values\.yaml']}" | base64 -d | yq '.tap_gui.app_config.app.baseUrl'
     ``` 
 
-1. (Optional) Create `api-auto-registration-values.yaml`
+1. (Optional but recommended) Create `api-auto-registration-values.yaml`
 
     If you'd like to overwrite the default values when installing the package, create a `api-auto-registration-values.yaml` file that looks like this:
 
+    This `cluster_name` field will be used as the lifecycle attribute for API entities created by the controller in TAP GUI.
```
LittleBaiBai, line 78: This line seems to be out of place. It's in between "file that looks like this" and the actual sample file. Aside from that, the description of this field is listed above with the sample `valuesSchema` output, so maybe we don't need to repeat it?

api-auto-registration/key-concepts.hbs.md
```diff
@@ -0,0 +1,105 @@
...
+        apiVersion:
+        kind:
+        name:
+        namespace:
+```
+
+Many of the above fields will result in specific behavior within TAP GUI.
+- The system and owner are copied to the API entity created. You may have to separately create and add to the catalog the [System](https://backstage.io/docs/features/software-catalog/descriptor-format#kind-system) and [Group](https://backstage.io/docs/features/software-catalog/descriptor-format#kind-group) kind. 
+- The namespace where the APIDescriptor CR is applied gets used as the namespace for API entity in TAP GUI. This results in the API entity's name, system and owner to all be under that namespace.
+- To explicitly use a system or owner in a different namespace, you can specify that in the respective field `system my-namespace/my-other system` or `owner: my-namespace/my-other-team`.
```
LittleBaiBai, line 34:
```suggestion
- To explicitly use a system or owner in a different namespace, you can specify that in the respective field `system: my-namespace/my-other-system` or `owner: my-namespace/my-other-team`.
- If the system or owner you are trying to link doesn't have explicit namespace specified, you can qualify them with `default` namespace, e.g.: `system: default/my-default-system`
```

api-auto-registration/key-concepts.hbs.md
```diff
@@ -0,0 +1,105 @@
...
+    path: "/v3/api-docs.yaml"
+    baseURL:
+      url: https://myservice.com
+```
+
+### <a id='with-ref'></a>With an Object Ref
+
+You can also use an object reference instead of hard coding the url. This can point to a HTTPProxy, Knative Service, or Ingress.
+
+Below is an example yaml that points to an HTTPProxy from which our controller extracts the `.spec.virtualhost.fqdn` as the baseURL.
```
LittleBaiBai, line 60: Could you add section headers with anchor for these different types of object refs? That way we can copy link to the specific section when sharing docs around.

api-auto-registration/troubleshooting.hbs.md
```diff
@@ -0,0 +1,21 @@
+# Troubleshooting
+
+Here are some handy commands for debugging or troubleshooting the APIDescriptor CR.
+
+1. Find the status of the APIDescriptor CR.
```
LittleBaiBai, line 5: The `-owide` flag is useful in some cases too. Could you add that as the first tip?

api-auto-registration/usage.hbs.md
```diff
@@ -1,14 +1,30 @@
 # API Auto Registration Usage
 
-## <a id='usage'></a>Using App Accelerator Template
+On a high level, API Auto Registration requires users to perform two steps: 1. generate API spec 2. create APIDescriptor CR.
```
LittleBaiBai, line 3: > generate API spec 

There are two paths here. One is to generate and expose API spec from the application, the other one is generate and publish API spec to a static location. Not sure what 's the best way to word it in this section, but we should keep this consistent as what we discussed in the key-concept section.

> requires users to perform

This makes the two steps sound like an active actions rather than something they can get for free. Maybe consider changing to something like: `requires two prerequisites: 1. somewhere to retrieve API spec from; 2. APIDescriptor CR getting created in the run clusters`?

api-auto-registration/usage.hbs.md
```diff
@@ -1,14 +1,30 @@
...
+   - [in an existing Spring Boot project using springdoc](https://springdoc.org/#getting-started)
+
+How to create APIDescriptor CR
+   - [using Out Of The Box Supply Chains](#using-ootb-supply-chain)
+   - [using Custom Supply Chains](#using-custom-supply-chain)
+   - [using other GitOps processes or Manually](#using-gitops-manually)
+
+How to configure
+   - [different cluster name, TAP GUI url or CA Cert values for api-auto-registration](#update-values)
+   - [CORS for OpenAPI Spec](#cors)
```
LittleBaiBai, line 17:
```suggestion
   - [CORS for viewing OpenAPI Spec in TAP GUI](#cors)
```

api-auto-registration/usage.hbs.md
```diff
@@ -83,75 +99,44 @@ spec:
...
-  system: spring-petclinic
-  owner: team-petclinic
-  location:
-    path: "/v3/api-docs.yaml"
-    baseURL:
-      url: https://myservice.com
-```
-
-### <a id='with-ref'></a>With a Ref
+## <a id='update-values'></a>Update install values for api-auto-registration package
```
LittleBaiBai, line 116: This section is almost exactly the same as the install guide. Could we just point them there?

Another option for updating install value is to update tap-values.yaml then update TAP. Not sure how other teams document that, maybe worth following the same pattern.

toc.md
```diff
@@ -165,6 +165,12 @@ docs.vmware.com is built.
         - [Additional resources](authn-authz/additional.md)
             - [Additional resources](authn-authz/additional.md)
             - [Install default roles independently](authn-authz/install.md)
+    - [API Auto Registration](api-auto-registration/about.md)
```
LittleBaiBai, line 168: Should we update the name of the product to `Tanzu API Auto Registration` to fit into the spot (like our neighbor `Tanzu API portal`), or `API Auto Registration for VMware Tanzu` since that's our official name?

LittleBaiBai (changes requested):

api-auto-registration/about.hbs.md
```diff
@@ -2,4 +2,19 @@
...
+registration from origin workloads. You may also use API Auto Registration without supply chain automation by direction 
+applying an `APIDescriptor` CR to the cluster.
+
+## <a id='getting-started'></a> Getting Started
+
+Learn more about APIDescriptor CR and API entities in TAP GUI in the [Key Concepts section](key-concepts.md)
+
+For "iterate", "run" and "full" TAP cluster profiles, start at the [Usage section](usage.md)
+
+For other profiles, install the api-auto-registration package using the [Installation section](installation.md)
```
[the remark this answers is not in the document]
LittleBaiBai: I see, you have a separate section in `usage` to talk about update. I'm fine with that option too.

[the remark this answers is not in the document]
shrutiyer: Yeah I debated repeating, but since the command for update was slightly different anyway and users have a focused page, I decided to go with the repeats. But I can check what other teams are doing

api-auto-registration/installation.hbs.md
shrutiyer: Same as previous. I will try to find the pattern that other teams follow. I think we have 3 scenarios, 1. full/iterate profile only optionally need to know about updates. 2. run requires the update. 3 all other need all the instructions. I may divide the instructions into sub sections to easily link to the concerned section for all these profiles.

api-auto-registration/installation.hbs.md
shrutiyer: You are right, I will remove it

shrutiyer, line 78:
```suggestion
```

toc.md
shrutiyer: I can change it to `API Auto Registration for VMware Tanzu` if that works. @csterwa any opinions?

api-auto-registration/about.hbs.md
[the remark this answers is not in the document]
shrutiyer: Check my newer commit's usage page first paragraph

api-auto-registration/installation.hbs.md
shrutiyer: Check my newer commit's usage page first paragraph

api-auto-registration/usage.hbs.md
shrutiyer: Updated. Check it out

jrshaffe (approved):

toc.md
LittleBaiBai: Maybe changing `Tanzu API portal` to `API portal for VMware Tanzu` as well? I don't know how we ended up with that name in the first place, but we can at least keep the names consistent for our two products.

LittleBaiBai (approved):

api-auto-registration/usage.hbs.md
LittleBaiBai: Looks good! Thanks! (I don't see the option to resolve this conversation but you can consider this one resolved.)

api-portal/about.hbs.md
```diff
@@ -1,5 +1,5 @@
-# Tanzu API portal
+# API portal for VMware Tanzu
 
 API portal for VMware Tanzu enables API consumers to find APIs they can use in their own applications. Consumers can view detailed API documentation and try out an API to see if it can meet their needs. API portal assembles its dashboard and detailed API documentation views by ingesting OpenAPI documentation from the source URLs. An API portal operator can add any number of OpenAPI source URLs in a single instance.
```
csterwa, line 4: Nice catch. I hadn't even identified this when reading through these.

csterwa (approved):

PR #1591, opened 2022-09-23
Update ALV architecture page for TAP 1.3

ksankaranara-vmw:
Only for TAP 1.3
Thanks.


RichardJJG:
Thanks, @ksankaranara-vmw

PR #1593, opened 2022-09-23
document windows support for IntelliJ

suarezjulian:
Which other branches should this be merged with (if any)? No, support for windows was only introduced in Tap 1.3


RichardJJG:
Thanks, @suarezjulian

PR #1597, opened 2022-09-26
Clarify the way dependsOn works

ericbottard:
Ref https://github.com/pivotal/acc-release/issues/377

This should be backported to all previous versions


RichardJJG:
Thanks

PR #1599, opened 2022-09-26
Draft: Jenkins Adapter documentation

rszumlakowski:


RichardJJG:
Thanks, @rszumlakowski, which other branches should a writer merge this PR with (if any)?

rszumlakowski:
This documentation is for TAP 1.3 and does not need to be back-ported.

PR #1602, opened 2022-09-26
PSTAR-799 viewing cluster with version >=1.24

rbentaarit:
@andreizimin 
@dimitrybesson


RichardJJG:
Thanks, @rbentaarit. Should I copy this to earlier live versions of the TAP docs, such as v1.2?

rbentaarit:
We support Kubernetes 1.24 as of TAP 1.3 so it is not necessary to include this in other releases.

PR #1606, opened 2022-09-27
Fix `imgpkg` tool name

evankanderson:

None


RichardJJG:
Thanks, @evankanderson

PR #1610, opened 2022-09-27
Update Resource Requirements

vdesikanvmware:
We have to update the resource requirements based on the findings on TAP 1.3.0.


RichardJJG:
Thanks, @vdesikanvmware

PR #1612, opened 2022-09-27
Exclude image policy webhook

elfotografo007:
This change is intended for TAP 1.3.0 and 1.2.2.


RichardJJG:
Thanks, @elfotografo007

PR #1613, opened 2022-09-27
PSTAR-866 updated TAP custom CA cert config and verif

rbentaarit:
1.3 +
@dimitrybesson 
@andreizimin


RichardJJG:
Thanks, @rbentaaritm. I changed `ingest` to `process`. If `ingest` is a special term that we should keep, let me know and I'll change it back.

PR #1617, opened 2022-09-28
Update Airgapped section for App Live View

ksankaranara-vmw:
Only for TAP 1.3.0. Thanks


RichardJJG:
Thanks, @ksankaranara-vmw

PR #1619, opened 2022-09-28
Add known issue - Tanzu CLI EKS connection failure

heyjcollins:

TAP 1.3 only.


RichardJJG:
Thanks, @heyjcollins

PR #1620, opened 2022-09-28
update AWS EKS connection failure section

heyjcollins:

This change can be included in TAP 1.2 and above.


RichardJJG:
Thanks, @heyjcollins

PR #1621, opened 2022-09-28
Insight/Store: Fix Service Account creation yaml indentation

garrying:
Signed-off-by: Garry Ing <ging@vmware.com>

Fixes an issue with the `annotations` key in the yaml not being nested under `metadata`. This is for kubernetes 1.24+.

cc @miketeve @Manifaust


RichardJJG:
Thanks, @garrying, does this apply to any earlier TAP docs versions also?

garrying:
@RichardJJG It doesn't apply to previous docs versions. Annotations are new with 1.3 to support kubernetes 1.24

PR #1630, opened 2022-09-29
updated TLK known issue.

mdc81:
Known issue https://jira.eng.vmware.com/browse/TANZUSC-1738 updated and added workaround.


HenryBorys:
@mdc81 We should probably add the issue with the "/" at the end of the key shared.image_registry.project_path to troubleshooting. But to do so, I need a little more context. Is this only a problem in 1.3? Do you expect it to be fixed sometime, or will it just be an ongoing issue? How would you describe the context. That is, when does it come up? What is the user trying to do when they experience this problem, and how does the problem show itself? What fails? Thanks.

mdc81:
@HenryBorys This will be a problem only in 1.3. No prior versions have this issue. And fix would be available in 1.3.1 onwards. Could you please review the Jira https://jira.eng.vmware.com/browse/TANZUSC-1738 and let me know if that has all the info that you are looking for? If not, I will try to eloborate.

PR #1631, opened 2022-09-29
Update App Live View pages for 1.3.0

ksankaranara-vmw:
Only for TAP 1.3.0


snneji:
Thanks @ksankaranara-vmw

PR #1638, opened 2022-09-30
Remove the Migrate from 1.3.0 Docs

mdc81:
I am not sure on how to remove all the references to migrate documentation from TAP Docs. Hence creating this PR to request for the removal.

Docs for 1.2 as well should not have the migration documents. It was supposed only in TAP 1.1 when the new profiles were introduced to support for the customers to migrate from light to iterate profile after an upgrade from TAP 1.0 to 1.1.


YugangW:
@mdc81 It seems [Light profile](https://docs-staging.vmware.com/en/draft/VMware-Tanzu-Application-Platform/1.3/tap/GUID-install.html#light-profile-3) section is the only place that is referencing the migrate topic. Do you want me to remove the entire Light profile section for both TAP 1.2 and 1.3 docs?

PR #1641, opened 2022-09-30
Updating the TLKs into build profile

mdc81:
Updating the TLKs into build profile as well.

### DIFF
--- a/multicluster/reference/tap-values-build-sample.hbs.md
+++ b/multicluster/reference/tap-values-build-sample.hbs.md
@@ -5,19 +5,30 @@ The following is the YAML file sample for the build-profile:
 ```yaml
 profile: build
 ceip_policy_disclosed: FALSE-OR-TRUE-VALUE # Installation fails if this is not set to true. Not a string.
+  
 shared:
+  ingress_domain: "INGRESS-DOMAIN"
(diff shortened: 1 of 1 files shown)


YugangW:
`ceip_policy_disclosed: FALSE-OR-TRUE-VALUE # Installation fails if this is not set to true. Not a string.` appears twice in line 7 and line 17. Do you want to keep both of them? @mdc81

multicluster/reference/tap-values-build-sample.hbs.md
```diff
@@ -5,19 +5,30 @@ The following is the YAML file sample for the build-profile:
...
+  ingress_domain: "INGRESS-DOMAIN"
   kubernetes_distribution: "openshift" # To be passed only for Openshift. Defaults to "".
-buildservice:
+  image_registry:
+    project_path: "SERVER-NAME/REPO-NAME" # Will be used by Build Service by appending "/buildservice" and used by Supply chain by appending "/workloads"
+    username: "KP-DEFAULT-REPO-USERNAME"
+    password: "KP-DEFAULT-REPO-PASSWORD"
+
+ceip_policy_disclosed: FALSE-OR-TRUE-VALUE # Installation fails if this is not set to true. Not a string.
+
```
divmadan, line 18: @YugangW @mdc81 
```suggestion
```

PR #1653, opened 2022-10-03
Review changes for Jenkins Adapter

rszumlakowski:


RichardJJG:
Should a writer cherry-pick this to any earlier versions of the docs, @rszumlakowski?

rszumlakowski:
Jenkins Adapter is for TAP 1.3 +.  It does not need to be backported to TAP < 1.3.

PR #1656, opened 2022-10-03
add aws-cli min supported version

heyjcollins:

TAP v1.2 and above.


heyjcollins:
Closing this PR in lieu of a different one which does a better job #1657

PR #1662, opened 2022-10-03
fix td-1012: remove incorrect statement wrt Live update

kdvolder:

Anything applying to 1.3 and higher.


kdvolder:
Related Jira: https://jira.eng.vmware.com/browse/TD-1012

PR #1669, opened 2022-10-03
Update more information on airgapped install

DennyHoang:

For TAP 1.3.0+


RichardJJG:
Thanks, @DennyHoang

PR #1670, opened 2022-10-03
TD-942: initellij monorepo

kdvolder:

1.3 and later


kdvolder:
Related Jira: https://jira.eng.vmware.com/browse/TD-942

kdvolder:
Some more details on how to actually use / setup this feaure might be required. But we can do that in a future / separate PR.

RichardJJG:
Thanks, @kdvolder

PR #1671, opened 2022-10-04
Vscode/1.3.0

kiwi-bui:
Which other branches should this be merged with (if any)? None, only latest.


RichardJJG:
Thanks, @kiwi-bui

PR #1672, opened 2022-10-04
Add Shared TLK in Update install-air-gap.hbs.md

dhadukk:
NA


RichardJJG:
Thanks, @dhadukk

PR #1688, opened 2022-10-05
Update accelerator configuration section

trisberg:

This is for TAP 1.3


RichardJJG:
Thanks, @trisberg

PR #1689, opened 2022-10-05
Update TAP-GUI RRV Metrics server workaround instructions

krisapplegate:
Applies only to TAP 1.3


tap-gui/plugins/runtime-resource-visibility.hbs.md
```diff
@@ -14,11 +14,34 @@ Do one of the following actions to access the Runtime Resources Visibility plug-
...
+## <a id="metrics-server"></a> If You Have a Metrics Server
 
 By default, the Kubernetes API does not attempt to use any metrics servers on your clusters.
 To access metrics information for a cluster, set `skipMetricsLookup` to `false` for that
-cluster in the `kubernetes` section of `app-config.yaml`.
+cluster in the `kubernetes` section of `app-config.yaml`. Below is an 
+
+```
+tap_gui:
+  ... // Existing configuration
```
mstergianis, line 25: My preference would be that we don't give users yaml with syntax errors. Even in docs. My suggestion would be to do something like
```yaml
tap_gui:
  # ... existing configuration
  app_config:
    # ... existing configuration
    kubernetes:
```
and so on

tap-gui/plugins/runtime-resource-visibility.hbs.md
```diff
@@ -14,11 +14,34 @@ Do one of the following actions to access the Runtime Resources Visibility plug-
...
+              authProvider: serviceAccount
+              serviceAccountToken: KUBERNETES_SERVICE_ACCOUNT_TOKEN
+              skipTLSVerify: true
+              skipMetricsLookup: false
+```
+
+Where:
+
+- `KUBERNETES_SERVICE_HOST` and `KUBERNETES_SERVICE_PORT` are the URL and ports of your Kubernetes cluster. These can be gathered via `kubectl cluster-info`.
+- `KUBERNETES_SERVICE_ACCOUNT_TOKEN` is the token from your tap-gui-token-id. You can retrieve this secret's ID  from `kubectl get secrets -n tap-gui` and then `kubectl describe secret tap-gui-token-ID` where ID is the secret name from the first step.
```
mstergianis, line 43: Should this maybe refer the customer to other docs that we give them on this? I think this explanation is missing steps (such as base64 decoding?) and instead of replicating the docs we might be able to say something like "you can retrieve this secret by following the docs here"

krisapplegate: Yeah I thought about this from [here](https://docs-staging.vmware.com/en/draft/VMware-Tanzu-Application-Platform/1.3/tap/GUID-tap-gui-cluster-view-setup.html) but it has a SA creation step fused in. I was going for the easy win. I'll defer to the docs team though.

tap-gui/plugins/runtime-resource-visibility.hbs.md
krisapplegate: Love it. I was looking for that syntax and to see if there was a standard way we were already doing that.

RichardJJG:
Thanks, @krisapplegate

PR #1692, opened 2022-10-05
Update about-package-profiles.hbs.md

sheltoG:
Added check for NodeJS functions.


RichardJJG:
Is this just for v1.3 docs, @sheltoG?

PR #1693, opened 2022-10-05
Create airgap-functions

sheltoG:
Adding an section for functions airgap setup.


RichardJJG:
@sheltoG, is this just for the v1.3 docs?

RichardJJG:
It's just for the v1.3 docs

PR #1696, opened 2022-10-05
Improve docs according to feedback

elfotografo007:
Signed-off-by: Andrés Torres <andrest@vmware.com>

Targeted for 1.3.0 and 1.2.2.


RichardJJG:
Thanks, @elfotografo007

PR #1700, opened 2022-10-05
Remove metadata store ingress values from example for Air gap install

Manifaust:
Because Store takes its values from shared TLK

This is related https://jira.eng.vmware.com/browse/TANZUSC-2045

Merge into main branch, this only applies 1.3


RichardJJG:
Thanks, @Manifaust

PR #1714, opened 2022-10-06
Update links to use new accelerator repository

eleftherias:
This commit also changes the examples to use tanzu-java-web-app instead of spring-petclinic, because spring-petclinic is not part of the accelerator repository.

This is only for 1.3


RichardJJG:
@eleftherias, does this apply to TAP v1.3.1, too? I ask because, if so, a writer will need to cherry-pick this to the `1-3-1` branch. Our branches are listed in the [README](https://github.com/pivotal/docs-tap#branches).

PR #1719, opened 2022-10-06
Update accelerator link for Services usecase

Samze:
Which other branches should this be merged with (if any)? TAP 1.3 and main


RichardJJG:
Thanks, @Samze

PR #1721, opened 2022-10-06
scc: add instructions configuring required SecurityContextConstraints for kaniko on OpenShift

cirocosta:
## Proposed changes

- add entry to the Dockerfile builds section detailing how to workaround the extra strictness that OpenShift places on running containers that require capabilities beyond the default unprivileged SecurityContextConstraints.

the changes are related to 1.3.x+


RichardJJG:
Thanks, @cirocosta

PR #1695, opened 2022-10-05
Draft: Adds back in TAP GUI app_config key in install

mstergianis:
It got removed some time ago when we went down to fewer required values.

Signed-off-by: Michael Stergianis <mstergianis@vmware.com>

Which other branches should this be merged with (if any)?

PR #1722, opened 2022-10-06
Add release notes for TAP GUI 1.3.2

weresch:
[PSTAR-784]

None, just for TAP 1.3


weresch:
Added app_config: to install.hbs.md to replicate #1695 as Michael is OOO and this change should be in for TAP 1.3 doc freeze.

RichardJJG:
Thanks, @weresch

PR #1723, opened 2022-10-06
VSEE-753 Add additional scan policy examples to include detail and ex…

nedenwalker:
…piration.


RichardJJG:
Thanks, @nedenwalker, is this just for TAP v1.3?

nedenwalker:
probably safer in 1.3 only.

PR #1730, opened 2022-10-07
updated to show the supported K8s versions to include v1.24

mdc81:
updated to show the supported K8s versions to include v1.24

1.3 Branch should have this change.


RichardJJG:
Thanks, @mdc81, we'll review and edit

PR #1733, opened 2022-10-07
fix: copy and paste errors

mingxiao:

This is for 1.3 and 1.2


RichardJJG:
@mingxiao, it's not applicable to v1.1 as well?

mingxiao:
@RichardJJG ah yes, this is a issue in 1.1 as well. Thank you!

PR #1734, opened 2022-10-07
Add note regarding misleading error

idoru:


RichardJJG:
@idoru, please resolve the conflict and then a writer will merge

PR #1739, opened 2022-10-07
fix: we do not support windows prior to TAP 1.3.0

mingxiao:

1.1 and 1.0


RichardJJG:
Thanks, @mingxiao

PR #1646, opened 2022-09-30
Update dependencies.hbs.md

johnnyr0x:
Updated the dependencies for TAP 1.3. Added jammy stacks.

PR #1740, opened 2022-10-07
Update buildpack versions for tap 1.3

tylerphelan:

this is TAP 1.3 only

replaces https://github.com/pivotal/docs-tap/pull/1646


snneji:
Thanks @tylerphelan!

PR #1741, opened 2022-10-07
Make configuring the tbs credentials as secret references more visibl…

tylerphelan:
…e and clear

- This addresses concerns about putting plaintext credentials in tap values.

This can be added to the 1.2 docs


RichardJJG:
Thanks, @tylerphelan

PR #1746, opened 2022-10-07
Remove more references to installing telemetry with a profile

xtreme-shane-lattanzio:
Which other branches should this be merged with (if any)? This change is from all releases before and including 1.3


RichardJJG:
Thanks, @xtreme-shane-lattanzio

PR #1749, opened 2022-10-07
scc: add reference page listing all scc parameters

cirocosta:
A lot of the parameters that apply to the underlying objects created according to the supplychains are scattered across multiple sections - this commit attempts at centralizing them in a single page such that one can more easily find them according to the difference resources that the most complete supply chain (ootb-supply-chain-testing-scanning) has.

jira: https://jira.eng.vmware.com/browse/TSCC-68


RichardJJG:
Thanks for putting the thought into this, @cirocosta

PR #1750, opened 2022-10-07
Add troubleshooting guidance for GUI error message about loading metadata store

Manifaust:
Also I added better descriptions when linking to how to enable CVE scan results.

I'm using partials to put the troubleshooting instructions in two different places, on in the TAP GUI docs and one in Store docs

Please merge into main as well as 1-2-2. This is because this troubleshooting guidance applies to both 1.2 and 1.3.

CC @miketeve


Manifaust:
This is also me trying to use partials and images at the same time so I hope that works as intended.

RichardJJG:
Thanks for this and the explanation, @Manifaust. I'll also have to see how we handle images within partials.

mdc81:
@Manifaust shouldnt this be listed in [Release Notes ](https://docs-staging.vmware.com/en/draft/VMware-Tanzu-Application-Platform/1.3/tap/GUID-release-notes.html) as well, from where one can access the corresponding troubleshooting steps?

RichardJJG:
> @Manifaust shouldnt this be listed in [Release Notes ](https://docs-staging.vmware.com/en/draft/VMware-Tanzu-Application-Platform/1.3/tap/GUID-release-notes.html) as well, from where one can access the corresponding troubleshooting steps? If not, where would one look for known issues?

My reading of the PR is that @Manifaust has provided troubleshooting steps for a user error (skipping a Getting Started section). We wouldn't describe user errors (even common ones) as known issues with the product. If this really is a product bug, @Manifaust, let me know and a writer will call it out in the release notes (hopefully before we publish).

PR #1765, opened 2022-10-11
do not use knative labels

mingxiao:
Use cartographer and kubernetes labels

This is for 1.3.0 and 1.3.1 only


RichardJJG:
@mingxiao, it's also for TAP v1.4 now, right? (`main` is now TAP v1.4 docs content)

mingxiao:
@RichardJJG yes!

PR #1774, opened 2022-10-12
Add missing pipe(|) after ca_cert_data

dhadukk:
only 1.3


RichardJJG:
@dhadukk, the `main` branch is now just for TAP v1.4 content. Do you want this change in TAP v1.4 and v1.3?

PR #1776, opened 2022-10-12
Attempt to pass links into a partial

Manifaust:
@RichardJJG I'm trying to get relative links to work, can you try this and let me know if it works?

### DIFF
--- a/partials/tap-gui/_ts-err-load-metadata-store.hbs.md
+++ b/partials/tap-gui/_ts-err-load-metadata-store.hbs.md
@@ -16,7 +16,7 @@ Supply Chain Security Tools - Store.
 #### Solution
 
 See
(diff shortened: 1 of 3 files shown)


RichardJJG:
Thanks for this. You want relative links to avoid the overhead of updating the version numbers in absolute links, right? 

We've found that we can [use variables for version numbers within absolute links](https://github.com/pivotal/docs-tap/blame/f1028d8666210e28a846fe4eb4b4781befbd1470/partials/_view-package-config.hbs.md#L62).

This variable method looks like the least amount of work going forward. Do you agree?

RichardJJG:
@Manifaust, sorry, I linked to the wrong thing.
This is [what I'm talking about](https://github.com/pivotal/docs-tap/commit/e0449311948dc258d69cb8d107b8e777ec2fdc85).

Manifaust:
@RichardJJG Yeah I think that could work, although it doesn't solve the issue of images.

Also, if you're going this route, you might consider turning the host name into a variable as well. So you can change `https://docs.vmware.com` to `https://docs-staging.vmware.com` more easily.

RichardJJG:
Ahh, interesting, thank you. Looping in @YugangW.

PR #1780, opened 2022-10-17
Fixed image link and note indentation

miketeve:
Main image was broken, fixed notes indentation, and updated note about how CVEs may not match

Which other branches should this be merged with (if any)? None - only 1.3


RichardJJG:
Thanks, @miketeve -- just to check, this applies to TAP v1.4, too, right? Your comment says "only 1.3", but the `main` branch is TAP v1.4 content now.

PR #1781, opened 2022-10-17
AppSSO: add docs for configuring AuthServer storage.

ddubson:
Which other branches should this be merged with (if any)? TAP 1.4 only.

Feature targeted for TAP 1.4 release.


RichardJJG:
Thanks, @ddubson

PR #1783, opened 2022-10-17
Refactor multicluster store docs

Manifaust:
Create a dedicated document for how to enable the multicluster deployment of Store. Link to it from the TAP multicluster docs.

This applies to main (1.4), 1.3, and 1.2.


RichardJJG:
Thanks, @Manifaust

PR #1784, opened 2022-10-18
Added imgpkg version query

krisapplegate:
Applies to docs 1.2+


RichardJJG:
Thanks, @krisapplegate

PR #1785, opened 2022-10-18
scc: add entry reference for deliverable parameters

cirocosta:
Similarly to how we have a reference for the parameters that can be set for workloads, here I introduce a section for Deliverable objects such that folks willing to customize some of the behaviors associated with the objects that get instantiated as a result of applying a `Deliverable` to a cluster that has `ClusterDelivery` installed (from `ootb-delivery-basic`) know what their options are.


a-peek4:
Thanks @cirocosta

PR #1801, opened 2022-10-21
Improve instructions on store secrets and namespaces

Manifaust:
- Also remove metadata store from iterate sample because store is not part of iterate
- Also better caveat `app_service_type` store configuration
- Describe how to export the metadata store CA to all namespaces
- Add `grype` config to the Install AWS sample tap values
- Better explain `ns_for_export_app_cert` in all the install docs

Merge this into main, as well as 1.3, and 1.2


RichardJJG:
Thanks, @Manifaust

PR #1803, opened 2022-10-24
Add EKS airgap config

DennyHoang:

please merge to main and 1-3-1


RichardJJG:
Thanks, @DennyHoang

RichardJJG:
@DennyHoang, in future, please just PR these sorts of changes to `main` (so that the writer only cherry-pick backwards). It's the safest way to change the docs.

PR #1804, opened 2022-10-24
Separate relocation and add repo

Petahhh:
* Main


RichardJJG:
Thanks, @Petahhh. In future, please PR changes like this to `main` only so that the writer only cherry-picks backwards. It's the safest way to edit the docs because we publish the production docs on a daily basis.

RichardJJG:
@Petahhh, I just realized I implied `1-3-1` is a production branch. It's not (yet) but my point still stands =P

Petahhh:
Hi @RichardJJG ! Thank you for taking the time to review, merge and cherry pick my PR/commit! Yes absolutely, I can PR to main and indicate afters the other branch it may apply to =)

PR #1807, opened 2022-10-24
VSEE-540 | Changing field term from 'failed' to 'errored' to be consistent with scans terminology

N00BEST:
**Which other branches should this be merged with (if any)?**
- This MR is intended to **Main**
- Please cherry pick to **1-3-1**


RichardJJG:
Thanks, @N00BEST

PR #1813, opened 2022-10-26
Create independent doc for Store custom certificate configuration

Manifaust:
Please merge this into `main`.

I will create another PR for the other branches.


RichardJJG:
@Manifaust, please don't create duplicate PRs for other branches -- it's ultimately less efficient than waiting for the writer to cherry-pick because of the extra work it makes.

RichardJJG:
I've learned this needs cherry-picking to `1-3-1`, `1-3-0`, and `1-2-2`.

RichardJJG:
The Tracker [story](https://www.pivotaltracker.com/story/show/183642025) is here.

Manifaust:
Thank you.

When merging into `1-3-1`, `1-3-0`, and `1-2-2`, the only issue I see is that there is a link that exist on `main` that will not be in the other older branches.

That link is for `scst-store/tls-configuration.hbs.md`, a page that exists in `main` but does not exist in the other places.

Another way to look at it is. There's a section of documentation inside `tls-configuration.hbs.md` that I have extracted out into its own docs `custom-cert.hbs.md`. This new doc is good because not only can I use it in `main`, but also backport it to the older branches.

RichardJJG:
Ah, thanks for the extra context

PR #1814, opened 2022-10-26
1.3 and 1.2 independent custom cert

Manifaust:
This is a copy of #1813 except it's meant to be merged into the 1.3.x and 1.2.x branches.

Please merge into `1-3-1` and `1-3-0` and `1-2-2`.

### DIFF
--- a/scst-store/additional.hbs.md
+++ b/scst-store/additional.hbs.md
@@ -12,4 +12,5 @@ This section includes the following topics:
 - [Troubleshooting upgrading](upgrading.md)
 - [Failover, redundancy, and backups](failover.md)
 - [Ingress support](ingress.hbs.md)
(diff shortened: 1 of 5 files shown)


Manifaust:
@RichardJJG sorry I just saw your other message

> please don't create duplicate PRs for other branches -- it's ultimately less efficient than waiting for the writer to cherry-pick because of the extra work it makes. Please just state here which other branches you want the content on.

I'll keep that in mind for next time.

The reason I created this PR is because there is one link that will cause a conflict. This link exists in 1.4 but does not exist in 1.3 and 1.2. So I thought I would help.

RichardJJG:
I've added the cherry-picking note to #1813 -- the writer knows to cherry-pick after she's made any corrections that the original PR needs. We have a lot of rules for house style, etc, and formatting constraints.

PR #1818, opened 2022-10-27
Update Store links in TOC and Additional

Manifaust:
Add .hbs extension to Store links in TOC and Additional page

Merge into `main`, `1-3-0`, `1-3-1`, and `1-2-2`.


Manifaust:
No I did not notice any problems with the functionality in the past.

This change was just to make clicking it inside my VSCode editor a bit easier.

HenryBorys:
Sounds good. I don't see any reason not to do this throughout the TOC for the same advantage in clicking links in VS Code.

HenryBorys:
In other words, I'll do that while I'm at it.

HenryBorys:
Not as simple as I thought, because of all the App-SSO content, so leaving that alone. But the changes @Manifaust made, look fine. Thanks.

PR #1823, opened 2022-11-01
Update multi-cluster View values

krisapplegate:
Applies to 1.2+


RichardJJG:
Thanks, @krisapplegate

PR #1826, opened 2022-11-02
Add docs and properties for using TLS for cluster internal calls in A…

trisberg:
…pp Accelerator

This is also needed for 1.3.1 docs.

Replaces #1490


RichardJJG:
Thanks, @trisberg. I'll warn now that we'll reverse the "username" and "repo" changes. House style dictates that "user name" is two words (although we'd never change a code reference, of course) and that we write out "repository" in full.

PR #1835, opened 2022-11-04
Tap gui/OIDC redirect url

mstergianis:
I think all of the branches feature this doc (and this info should be pertinent for all of them), so yeah, it should hopefully improve things for all of them :smile:


mstergianis:
@RichardJJG I sent this to you via a slack thread we have, but am also going to ping you here for convenience. I wish I could just assign people :shrug:

RichardJJG:
Thanks! I'll cherry-pick as far as I can.

PR #1841, opened 2022-11-07
Update prerequisites.hbs.md

alexbarbato:
An attempt to qualify the run as root requirements

Main

### DIFF
--- a/prerequisites.hbs.md
+++ b/prerequisites.hbs.md
@@ -132,8 +132,10 @@ For more information, see the [Kubernetes documentation](https://kubernetes.io/d
 
 - For the [`full` profile](install.html#full-profile) or use of Security Chain Security Tools - Store, your cluster must have a configured default StorageClass.
 
-- Pod security policies must be configured so that Tanzu Application Platform controller pods can run as root.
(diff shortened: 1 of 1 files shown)


RichardJJG:
Thanks, @alexbarbato. Does this really only apply to TAP v1.4 (`main`)? I ask because it sounds like information that has always been true -- or at least has been true for a long time.

RichardJJG:
I've read (some) of this [Slack thread](https://vmware.slack.com/archives/C02D60T1ZDJ/p1658276392011849). Thank you for investigating and please raise a new PR for the Kaniko reference and anything else you wish you add (I've already merged this PR, so it's easiest to just make new PR).
The [Google doc](https://docs.google.com/document/d/1AezJsCkiF3ZbwLyJqb3fX7yihjMgyQyES1ToemahGjc/edit#heading=h.6ruirzh64um2) suggests I need to cherry-pick all this to the v1.3 docs. Do you know if I need to go back even further than that?

alexbarbato:
I don't know that you need to go back further than that, but I think 1.3 is sufficient. May be worth asking in that thread though!

I'll make the new PR now

PR #1846, opened 2022-11-08
Draft: Add gitlab discovery processor to TAP GUI docs

weresch:
1-3-1 and main


RichardJJG:
Thanks, @weresch

RichardJJG:
For future reference, @weresch, we prefer that PRs such as these be for `main` instead of a production branch.

PR #1847, opened 2022-11-08
Add namespace to tanzu package installed list

weresch:
1-3-0, 1-3-1, main


RichardJJG:
We don't want this in the v1.2, v1.1, and v1.0 docs as well? It looks like it applies to me, but I could be wrong.

weresch:
@RichardJJG No, because the previous commands in that doc make no mention of the tap-install namespace

PR #1848, opened 2022-11-08
Document the developer conventions scc for openshift clusters

paulcwarren:
This is also for tap 1.3.x

Please let me know if I need to update any index pages.  I wasn't sure so I left it for now.  thx


RichardJJG:
Thanks, @paulcwarren

PR #1849, opened 2022-11-09
AppSSO: document `default_authserver_issuer`

mamachanko:
[#183569338]


YugangW:
Hi @mamachanko, Is this change for TAP 1.4 only? Thanks!

mamachanko:
> Hi @mamachanko, Is this change for TAP 1.4 only? Thanks!

@YugangW yes, it is!

PR #1851, opened 2022-11-09
add known issue

mingxiao:
Tanzu Debug does not port forward application port

Just 1.4. And this is a breaking change


RichardJJG:
Thanks, @mingxiao

PR #1852, opened 2022-11-09
Learning Center: Removing master references that are not required

marroyo:


RichardJJG:
Should a writer cherry-pick this to any other branches, @marroyo?

PR #1856, opened 2022-11-10
Inclusion updates for 1-2-2

marroyo:


RichardJJG:
Thanks, @marroyo, you don't want this cherry-picked anywhere, right?

PR #1866, opened 2022-11-10
Document 'UnsupportedClassVersion' workaround for vscode live-update

kdvolder:

This change applies to 1.3.1 and newer. So please apply change to 1.3.1 and main branch.


kdvolder:
For further background see this slack thread: https://vmware.slack.com/archives/C03GZ4LRTH6/p1667977844176369

RichardJJG:
Thanks, @kdvolder

PR #1872, opened 2022-11-14
AppSSO: Extend trust for CA per `AuthServer`

mamachanko:


RichardJJG:
When this is ready for merging, @mamachanko, which other branches (if any) do you want a writer to cherry-pick it to?

mamachanko:
> When this is ready for merging, @mamachanko, which other branches (if any) do you want a writer to cherry-pick it to?

@RichardJJG  This is for TAP `1.4.0`. I understand only `main` should be targeted then.

RichardJJG:
Because we're now in Docs Freeze, this possibly won't be merged until after November 15. That said, because this is for v1.4 only, that's obviously no big deal.

PR #1875, opened 2022-11-14
Document IntelliJ 'corrupted launch config' workaround

kdvolder:

All branches after 1.3.1 so I think that means 1-3-2 and main.

### DIFF
--- a/vscode-extension/troubleshooting.hbs.md
+++ b/vscode-extension/troubleshooting.hbs.md
@@ -55,3 +55,39 @@ You can force the VScode Java tooling to re-read and synchronize information fro
 
 This should trigger the internal compiler level to then be set correctly based
 on the information from the `pom.xml` (e.g. Java 11 in the `tanzu-java-web-app` example).
+
+## <a id='debug-not-working-corrupt-launch-conf'>Starting a Debug Session Errors with 'Unable to open debugger port'
+
+### Symptom
+
+You try to start a 'Tanzu Debug' session and it immediately fails an error like:
+
+```
+Error running 'Tanzu Debug - fortune-teller-fortune-service': Unable to open debugger port (localhost:5005): java.net.ConnectException "Connection refused"
+```
+
+### Cause
+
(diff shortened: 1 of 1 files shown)


kdvolder:
Related Jira: https://jira.eng.vmware.com/browse/TD-1064

a-peek4:
Hi @kdvolder We received your PR after the doc freeze date. We are currently working on release tasks, and we will review your PR as soon as we are able. Unfortunately, this may mean that your PR will not be published until after the release. Going forward, to ensure that your PRs are included on release day, please submit them no later than 23:59 UTC-8 on the [doc freeze date](https://confluence.eng.vmware.com/pages/viewpage.action?pageId=789919969).

kdvolder:
Sure, I understand, and thanks for doing whatever you can to get it in. If it gets in before release that is great. If it happens after release so be it, it really isn't a big deal. 

> please submit them no later than 23:59 UTC-8 on the [doc freeze date]

We will try, but really, the PR was submitted as soon as it was ready. While we may hope/try to put all changes in before a certain date, this can't always be done. 

I would assume that since docs are essentially just web-pages the content of which we can 'in theory' update at any time, it shouldn't really stop us from submitting changes after the 'freeze date'. Even after a product was released, I beleave we should be stll be able to update its docs as we discover the need to do so.

If that isn't the case and docs become 'read only' after release/freeze, then I can resubmit a PR for 1.3.2 branch (in fact I considered doing that, but when I look at 1.3.2 branch's content it looks like the various bits and pieces from 1.3.1 that need to be cherry picked may not yet be in that branch... so I decided instead to do the PR against 1.3.1 which seems to have most up-to-date content).

Let me know if need/want me to rebase PR against 1.3.2 branch if its too late for 1.3.1 doc changes.

RichardJJG:
@kdvolder, you're correct in that the docs are alive and subject to change until we archive them when the relevant product version leaves general support.
The Docs Freeze period exists because the writers need time to get through the last of the PR backlog and prepare to safely piublish.

PR #1883, opened 2022-11-15
Update services docs to make use of class claims

teddyking:
Class Claims are a new feature in TAP 1.4 and this commit updates the corresponding documentation to make use of them.

Which other branches do you want a technical writer to cherry-pick this PR to (if any)? None - these changes are only relevant from TAP 1.4 onwards.

### DIFF
--- a/getting-started/consume-services.hbs.md
+++ b/getting-started/consume-services.hbs.md
@@ -14,25 +14,25 @@ For more information about these prerequisites, see [Set up services for consump
 
 ## <a id="you-will"></a>What you will do
 
-- Inspect the resource claim created for the service instance by the application operator.
-- Bind the application workload to the ResourceClaim so the workload utilizes the service instance.
+- Inspect the claim created for the service instance by the application operator.
+- Bind the application workload to the claim so the workload utilizes the service instance.
 
 ## <a id="overview"></a>Overview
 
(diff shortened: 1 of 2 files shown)


teddyking:
Hi there, as part of this change we'll also need to update the corresponding diagrams on the "Set up services for consumption by developers" and "Consume services on Tanzu Application Platform" pages. How can we go about doing that? I don't think I have access to the source image for those diagrams 🤔 

I've attached a rough diagram depicting the changes to make:

updated-diagram-stk

* Change 1 - Update the text from `$ tanzu service claim create` to `$tanzu service class-claim create`
* Change 2 - Create a new box for `ClassClaim`. The green arrow from App Operator should now point to the `ClassClaim` and there should be green arrow from `ClassClaim` to `ResourceClaim`.

Everything else on the diagram should remain the same.

Thanks!

HenryBorys:
@teddyking Because we don't have the original, we need to submit a ticket to Content Ops for this. I can do it if you like.

teddyking:
@HenryBorys Ah ok, yes please that would be great, thanks!

HenryBorys:
Okay, I've put in the ticket: https://jira.eng.vmware.com/browse/CO-4163

HenryBorys:
@teddyking What versions of the docs will this updated image apply to?

teddyking:
Thanks @HenryBorys - this change only applies to TAP version 1.4+ so no need to backport it to earlier branches.

HenryBorys:
@teddyking Cool! Thanks!

PR #1885, opened 2022-11-16
Delete App Accelerator "Specs" files

ericbottard:
Those files are not meant to be published to end users.

=> ALL


RichardJJG:
Thanks, @ericbottard

PR #1886, opened 2022-11-16
Document custom types

ericbottard:
Fix https://github.com/pivotal/acc-release/issues/433

=> None (1.4 feature)


application-accelerator/creating-accelerators/custom-types.hbs.md
```diff
@@ -0,0 +1,163 @@
...
+for the `dataType` property of any option:
+```yaml
+accelerator:
+  options:
+    - name: myTask
+      dataType: Task
+  types:
+    ...
+```
+whereby UIs would render something similar to the following:
```
kcoriordan, line 48: I'm not clear what this means?
What UI is being referred to here?

application-accelerator/creating-accelerators/custom-types.hbs.md
```diff
@@ -0,0 +1,163 @@
...
+![SimpleTask](simple-task.png)
```
kcoriordan, line 49: This graphic has a mixture of languages. There is a different language under Enter task details
It would be better to get an English only graphic if you could supply one

application-accelerator/creating-accelerators/custom-types.hbs.md
```diff
@@ -0,0 +1,163 @@
...
+
+and associate the entered
```
kcoriordan, line 51: Is this a run on sentence from somewhere - can this be reworded - it's not very clear.

kcoriordan:
@ericbottard 
I have made some editorial updates to custom-types topic- here is [preview](https://docs-staging.vmware.com/en/draft/VMware-Tanzu-Application-Platform/1.5/tap/application-accelerator-creating-accelerators-custom-types.html)
I have added some inline questions above

application-accelerator/creating-accelerators/custom-types.hbs.md
ericbottard: That's the thing. All /  any UI. Hence the use of plural and "something like".
This doc is not about how exactly this is rendered by UIs, but rather the  feature of  having tabular data.
This particular screenshot is taken from something similar to the backstage UI
Hope that makes sense.

application-accelerator/creating-accelerators/custom-types.hbs.md
ericbottard: Yes it  is (it was, I saw you updated it).
Sadly the sentence doesn't make sense anymore, sorry.

There was a train of thoughts in the previous sentence.
It BECAUSE you'd use such yaml construct THAT uis would render such widget AND THAT the entered values would be associated with the example value submission. You've changed the subject of  the sentence (and lost the first BECAUSE/THEN structure).

kcoriordan:
OK @ericbottard - I've change those 2 lines back now, can you check the meaning is intact through out now, thanks
https://docs-staging.vmware.com/en/draft/VMware-Tanzu-Application-Platform/1.5/tap/application-accelerator-creating-accelerators-custom-types.html

PR #1889, opened 2022-11-17
Change CVEs resolution docs to VMware required format.

gmrodgers:
1.3.1 only is fine


RichardJJG:
Thanks, @gmrodgers

PR #1891, opened 2022-11-18
Add note of how tap-values is stored as a kubernetes secret

xyloman:
Which other branches do you want a technical writer to cherry-pick this PR to (if any)? n/a

Make the docs more clear that the tap-values.yaml is stored as a kubernetes secret.

cc: @miketeve


RichardJJG:
Thanks, @xyloman

PR #1927, opened 2022-11-28
Add docs for build-image-action

emmjohnson:


tanzu-build-service/build-image-action.md
```diff
@@ -0,0 +1,137 @@
...
+    # If not running locally, use `${{ secrets.CA_CERT }}`
+    ca_cert: ''
+    # Service Account token to access kubernetes
+    # If not running locally, use `${{ secrets.TOKEN }}`
+    token: ''
+    # _(required)_ The namespace to create the build resource in
+    namespace:
+
+    ## Image configuration
+    # _(required)_ Destination for the built image
```
zrob, line 94: is this required or optional? i thought tbs was configured with a default location for writing images which would make this optional?

also how do credentials for writing to for example the gcr registry get passed through in this case?

garethjevans: At the moment this is a required field, we could look at querying the value for `buildservice.kp_default_repository` and construct a sensible default for the repo.

Credentials are configured in the TAP cluster. From the github side, the only thing we allow the user to configure is the name of the serviceAccount used to perform the build, that way, the credentials never leave the TAP cluster.

emmjohnson: I added some info on setting up the developer namespace - https://github.com/pivotal/docs-tap/pull/1927/files#diff-157384f69e31eab7b1a9eaea17d411ebf1bb7bc5398263dbee515fe07b4cb639R15

kcoriordan:
Hi Emma, @emmjohnson 
I've made some style and structure updates to the new topics to bring it into line with our style standards - content should be unchanged. Can you take a look please to ensure it all still looks good.

PR #1934, opened 2022-11-30
update kubernetes_version TLK for openshift

vdesikanvmware:


snneji:
Hi @vdesikanvmware does this apply to any other branches?

PR #1941, opened 2022-12-01
appsso: add LDAP identity provider configuration

Kehrlann:
Only for TAP 1.4

### DIFF
--- a/app-sso/service-operators/identity-providers.md
+++ b/app-sso/service-operators/identity-providers.md
@@ -85,9 +85,7 @@ accessible user URI for the `AuthServer`, including scheme and port is `spec.iss
 accessible on `https://appsso.company.example.com:1234/`, the redirect URI registered with the identity provider should
 be `https://appsso.company.example.com:1234/login/oauth2/code/my-provider`.
 
-## LDAP (experimental)
-
-> *WARNING:* Support for LDAP providers is considered "experimental".
+## LDAP
 
 **At most one** `ldap` identity provider can be configured.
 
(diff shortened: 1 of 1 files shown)


app-sso/service-operators/identity-providers.md
```diff
@@ -103,24 +101,21 @@ spec:
...
+          searchFilter: cn={0}
         group:
-          searchFilter: member={0}
-          searchBase: ou=Users,o=5d03d6ac6eed091436a8d664,dc=jumpcloud,dc=com
-          searchSubTree: true
-          searchDepth: 10
-          roleAttribute: cn
+          search:
+            filter: member={0}
+            base: ou=Users,dc=example,dc=com
```
candicenonsense, line 115: @Kehrlann do we want to refer to this as memberOf to make it look more AD-ish?

Kehrlann: @candicenonsense Ah sorry, I missed this comment somehow.

`memberOf` is only used when `group.search` is turned off. For clarification, `memberOf` is an attribute of the user object, while `member` is an attribute of the group object.

YugangW:
Hi @Kehrlann, Groups to roles mapping seems like an empty section: https://github.com/pivotal/docs-tap/pull/1941/files#diff-1dbf534060853f91e91f0aaee6b44100649a80bc6900f89c07aece293b869b4aR145

Is this something by design? Thanks!

Kehrlann:
@YugangW good catch, that's actually an issue. Feel free to remove it entirely.

PR #1945, opened 2022-12-02
fix typo

mamachanko:
This change is targeted for TAP `1.4.0`

---

* none

* ack


RichardJJG:
Thanks,  @mamachanko

PR #1954, opened 2022-12-02
appsso: fix appsso > service ops > identity providers formatting

Kehrlann:
Fix formatting.

Only TAP 1.4


RichardJJG:
Thanks, @Kehrlann

PR #1973, opened 2022-12-07
Add API validation analysis related docs

shrutiyer:
@karinepires what is the link for Mazinger installation.

@karinepires @visomar @jrshaffe any edits or additions to make these sections more informative? Also please tag others who worked on these features

@mstergianis @krisapplegate @RichardJJG @YugangW also please add comments


jrshaffe (approved):

tap-gui/plugins/api-docs-getting-started.hbs.md
```diff
@@ -171,3 +171,59 @@ To update your demo API entry:
...
+```
+
+### <a id="automatic-validation"></a> Automatic OpenAPI specification Validation
+
+You can receive automatic validation analysis for OpenAPI specs using APIx.
+Follow the installation using the APIx docs. (TODO insert links to their doc).
+In addition, you will need to use either [API Auto Registration](../../api-auto-registration/about.hbs.md) or
+APIx Design GitOps to automatically generate the API entities in TAP GUI.
+
+Note: The automatic scoring will not be able to score or replace API entities created through other methods like regular GitOps or manual registration.
```
mstergianis, line 229: Suggestion: we note this but I think we should also tell users what it looks like when they encounter this. For what component should they check the logs? What do the logs look like?

Expanding on "for what component should they check the logs" maybe this should link to a section in another components docs where that explainer lives :shrug:. IIRC TAP GUI will report a HTTP error for the components that do not belong to the immediate entity provider, so maybe it's a joint effort between TAP GUI and API auto reg to provide docs to educate customers how to detect when this is happening

mstergianis (commented):
Looks mostly good to me :smile:

tap-gui/plugins/api-docs-getting-started.hbs.md
shrutiyer: The error message is not quite how you expect it to be. IIUC, the only "error message" (which is more like a regular message) inside the validation analysis card itself. Check out the latest changes for what I mean. Lmk what you think

mstergianis (approved):

RichardJJG:
I don't have comments at this stage. 
When the PR is factually accurate and complete to the standards of the SMEs, it leaves draft mode and a writer (likely me) merges it and edits it. I might have a clarifying question or two at that point, but you don't want my input earlier than that because it would be too many cooks in the kitchen.
Also, when the draft is still too rough to merge the SMEs have a better idea of what is lacking than I do.

RichardJJG:
Thanks, @shrutiyer, I've merged: you can create a new PR for that link whenever you like.

PR #1981, opened 2022-12-08
More updates

emmjohnson:
Which other branches do you want a technical writer to cherry-pick this PR to (if any)? no

### DIFF
--- a/tanzu-build-service/build-image-action.md
+++ b/tanzu-build-service/build-image-action.md
@@ -11,89 +11,124 @@ Use this GitHub Action to creates a Tanzu Build Service Build on a cluster.
 
 ## Procedure
 
+### Developer Namespace
+
 1. Create a developer namespace where the build resource will be created.
 
     ```bash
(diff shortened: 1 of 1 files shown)


garethjevans (approved):

tanzu-build-service/build-image-action.md
```diff
@@ -11,89 +11,124 @@ Use this GitHub Action to creates a Tanzu Build Service Build on a cluster.
...
     ```
 
-4. This action needs access to the cluster, using the Kubernetes service account created, with the
-following [GitHub encrypted secrets](https://docs.github.com/en/actions/security-guides/encrypted-secrets) set:
-
-    - CA_CERT
-    - NAMESPACE
-    - TOKEN
-    - SERVER
+Then to access the values (on GKE, if using another IAAS, this may be different):
```
kcoriordan, line 119: This is the command to access the values in GKE - right?
So it should say:
>To access the values on GKE (if using another IAAS, this may be different):

PR #2020, opened 2022-12-16
add Docker registry url fix example

heyjcollins:

this additional troubleshooting item can be backported as far back as this troubleshooting page has existed.

the `---` separators between each TS item really help with readability (to see where each TS item begins/ends), so I'd recommend making that change as well although, it would have to be manually added to older versions since they won't have all the same number of TS entries...


RichardJJG:
Thanks for this consideration wrt to horizontal lines, @heyjcollins

PR #2026, opened 2022-12-20
Add known issue for IntelliJ tools

kdvolder:
> Which other branches do you want a technical writer to cherry-pick this PR to (if any)?

This change only applies to the 1.4 release. So I think that means it only goes in `main` branch?

### DIFF
--- a/release-notes.hbs.md
+++ b/release-notes.hbs.md
@@ -204,7 +204,7 @@ This release has the following known issues, listed by area and component.
   activity pane. For more information, see
   [Troubleshooting](intellij-extension/troubleshooting.hbs.md#tnz-panel-k8s-rsrc-fail).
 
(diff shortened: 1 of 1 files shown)


release-notes.hbs.md
```diff
@@ -204,6 +204,10 @@ This release has the following known issues, listed by area and component.
   activity pane. For more information, see
   [Troubleshooting](intellij-extension/troubleshooting.hbs.md#tnz-panel-k8s-rsrc-fail).
 
+- The Details table and Messages pane in the activity panel nay show stale data. The reason
```
tyuzhikova-vmware, line 207: Typo in `may`

kdvolder: Yes :-) fixed. Thanks!

PR #2048, opened 2023-01-03
Adds a release note for TAP GUI around auto TLS

mstergianis:
Signed-off-by: Michael Stergianis <mstergianis@vmware.com>

None

### DIFF
--- a/release-notes.hbs.md
+++ b/release-notes.hbs.md
@@ -53,6 +53,41 @@ This topic contains release notes for Tanzu Application Platform v1.4.
 - `cert-manager.tap.tanzu.vmware.com` can optionally install self-signed `ClusterIssuer`s.
 </br></br>
 
+#### <a id="1-4-0-tap-gui-nf"></a> Tanzu Application Platform GUI
+
+- As mentioned in the [Tanzu Application Platform new features
+  section](#1-4-0-tap-new-features) Tanzu Application Platform GUI participates
+  in the shared ingress issuer feature. As such if you have explicitly set the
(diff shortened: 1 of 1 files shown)


andreizimin:
@mstergianis @RichardJJG - should we just focus on the "new" version - how it should look like?

mstergianis:
Not opinionated if we show both the old format and the new one, or just the new one.

Honestly I think my preference would be a diff format, something like
```diff
--- old-tap-values.yaml	2023-01-03 15:21:05.110897726 -0500
+++ new-tap-values.yaml	2023-01-03 15:21:17.147698490 -0500
@@ -1,8 +1,8 @@
 tap_gui:
   app_config:
     app:
-      baseUrl: http://tap-gui.INGRESS-DOMAIN/
+      baseUrl: https://tap-gui.INGRESS-DOMAIN/
     backend:
-      baseUrl: http://tap-gui.INGRESS-DOMAIN/
+      baseUrl: https://tap-gui.INGRESS-DOMAIN/
       cors:
-        origin: http://tap-gui.INGRESS-DOMAIN/
+        origin: https://tap-gui.INGRESS-DOMAIN/
```

So maybe I should have done that in the first place. Does our docs site have the ability to render a diff like it does yaml? That'd be nice :smile:

RichardJJG:
@mstergianis, @andreizimin 

> Does our docs site have the ability to render a diff like it does yaml?

No, but I've been asking for code syntax highlighting for a long time. There is a [JIRA issue](https://jira.eng.vmware.com/browse/CDF-20121) you can vote for.

> should we just focus on the "new" version - how it should look like?

I think that would be neatest, given the limitations.

mstergianis:
Excellent thanks @RichardJJG

PR #2050, opened 2023-01-03
Docs/accelerator updates 1 4

matt-dres:

1.4+

Additional context on changes:
* Revised version numbers where relevant - aiming to make them less release-version dependent for easier maintenance.
* Added `shell` compatible variables where relevant - aiming to improve the reader's ability to truly copy and paste our examples
* Added additional callout for custom types in `creating-accelerators.hbs.md`
* Added in additional examples for retrieving information regarding k8s secrets & the FQDN for TAP GUI
* Revised `YAML` examples to instead have variables use `_` instead of `-`


application-accelerator/creating-accelerators/creating-accelerators.hbs.md
```diff
@@ -180,10 +182,11 @@ spec:
   displayName: Select Java Version
   git:
     ref:
-      tag: tap-1.3
+      tag: TAP_VERSION
     url: https://github.com/vmware-tanzu/application-accelerator-samples.git
     subPath: fragments/java-version
 ```
+Where `TAP_VERSION` is the current version the Tanzu Application Platform. As an example, this will take the form: `tap-X.Y.Z`, where `X.Y.Z` are version numbers. (e.g. `tap-1.4`)
```
trisberg, line 189: should be: (e.g. `tap-1.4.0`)

application-accelerator/creating-accelerators/creating-accelerators.hbs.md
```diff
@@ -198,10 +201,12 @@ To avoid having to create a separate manifest file, you can use the following co
 ```
 tanzu accelerator fragment create java-version \
   --git-repo https://github.com/vmware-tanzu/application-accelerator-samples.git \
-  --git-tag tap-1.3 \
+  --git-tag ${TAP_VERSION} \
   --git-sub-path fragments/java-version
 ```
 
+Where `TAP_VERSION` is the `spec.git.ref.tag` value that is located in the `java-version.yaml`. (e.g. `tap-1.4`)
```
trisberg, line 208: This is confusing - the tag is the git repo tag used for specific versions of the samples that are delivered for each TAP release like `tap-1.3`, `tap-1.3.1` or `tap-1.4.0` - you can see the list here: https://github.com/vmware-tanzu/application-accelerator-samples/tags

application-accelerator/creating-accelerators/transforms/open-rewrite-recipe.hbs.md
```diff
@@ -3,8 +3,15 @@
 The `OpenRewriteRecipe` transform allows you to apply any [Open Rewrite](https://docs.openrewrite.org/)
 **Recipe** to a set of files and gather the results.
 
->**Note** Currently, only [Java-related recipes](https://docs.openrewrite.org/reference/dependency-version-selectors/java/)
-are supported. The engine leverages version `7.21.3` of Open Rewrite and parses
+>**Note** Currently, only the following Open Rewrite Recipes are supported:
```
trisberg, line 6: remove "only" since we support quite a few of them

application-accelerator/creating-accelerators/transforms/open-rewrite-recipe.hbs.md
```diff
@@ -3,8 +3,15 @@
...
+>  * [Java recipes](https://docs.openrewrite.org/reference/recipes/java)
+>  * [Maven recipes](https://docs.openrewrite.org/reference/recipes/maven)
+>  * [XML recipes](https://docs.openrewrite.org/reference/recipes/xml)
+>  * [YAML recipes](https://docs.openrewrite.org/reference/recipes/yaml)
+>  * [JSON recipes](https://docs.openrewrite.org/reference/recipes/json)
+>  * [Properties recipes](https://docs.openrewrite.org/reference/recipes/properties)
+>
+> The engine leverages version `7.21.3` of Open Rewrite and parses
```
trisberg, line 14: The current version is `7.30.1` but not sure we need to be that specific here since version used will change over time

application-accelerator/install-app-acc.hbs.md
```diff
@@ -75,21 +75,21 @@ To install Application Accelerator:
     $ tanzu package available list accelerator.apps.tanzu.vmware.com --namespace tap-install
     - Retrieving package versions for accelerator.apps.tanzu.vmware.com...
       NAME                               VERSION  RELEASED-AT
-      accelerator.apps.tanzu.vmware.com  1.3.1    2022-09-30 13:00:00 -0400 EDT
+      accelerator.apps.tanzu.vmware.com  1.4.0    2023-01-10 13:00:00 -0400 EDT
```
trisberg, line 78: The current value for 1.4.0 release is

```
  NAME                               VERSION  RELEASED-AT                    
  accelerator.apps.tanzu.vmware.com  1.4.0    2022-12-08 12:00:00 -0500 EST  
```

application-accelerator/install-app-acc.hbs.md
```diff
@@ -109,26 +109,37 @@ To install Application Accelerator:
...
       >   service_type: "ClusterIP"
       >   watched_namespace: "accelerator-system"
       > samples:
       >   include: true
-      >```.
+      > ```
+    
+    >**Note** When working instead with a `tap-values.yaml` file, ensure that the top level parameter is `accelerator:`. Using the example above as reference:
+      > ```yaml
+      > accelerator:
```
trisberg, line 122: using this `accelerator` entry won't work when installing the accelerator package individually - the properties must be at the top level in the yaml file

application-accelerator/vscode.hbs.md
```diff
@@ -4,7 +4,7 @@ The application Accelerator Visual Studio Code extension lets you explore and ge
 from the defined accelerators in Tanzu Application Platform using VS Code.
 
 ## <a id="dependencies"></a> Dependencies
-
+<!-- TODO Is this still needed for this section?? -->
```
trisberg, line 7: Needs to be updated for interacting with the tap-gui server endpoint

partials/application-accelerator/_configuration.hbs.md
```diff
@@ -206,17 +208,17 @@ Accelerator. The same properties are provided in the `accelerator` section of th
 accelerator:
   git_credentials:
     secret_name: git-credentials
-    username: GIT-USER-NAME
-    password: GIT-PASSWORD-OR-ACCESS-TOKEN
-    ca_file: CUSTOM-CA-CERT
+    username: GIT_USERNAME
```
trisberg, line 211: why use underscores?

trisberg:
For placeholders, we should follow the style that is used throughout the TAP docs

application-accelerator/creating-accelerators/creating-accelerators.hbs.md
matt-dres: I'll rephrase this to call out the git tag rather than the TAP version. Good call 👍

application-accelerator/creating-accelerators/transforms/open-rewrite-recipe.hbs.md
matt-dres: I removed mention of Open Rewrite in this line (didn't think it added much info if no version was there) and kept the portion on parsing Java files using Java 11 grammar.

partials/application-accelerator/_configuration.hbs.md
matt-dres: From looking at the K8s documentation, specifically their example YAMLs, they leverage underscores rather than dashes. At first, I was switching things over to using the form `$VARIABLE_NAME`, but wasn't too sure if it was too big of a jump to assume folks had `envsubst` available.

If this goes against the style for the TAP docs, can certainly revert.

application-accelerator/install-app-acc.hbs.md
matt-dres: Got it - so this addition should be entirely removed. Is that right?

application-accelerator/vscode.hbs.md
```diff
@@ -32,25 +32,33 @@ Use the following steps to install the Application Accelerator Visual Studio ext
...
+accessing the Application Accelerater extension icon:
 
 ![The explorer panel is open, the TIMELINE drop-down is selected, and the Demo Types icon is highlighted.](../images/app-accelerators-vscode-icon.png)
 
 Choose any of the defined accelerators, fill the options and click  the `generate project`
 
 ![The accelerator tab is open to the Hello Fun accelerator form. The text boxes display example text and the Generate Project button is highlighted.](../images/acc-form.png)
+
+## <a id="app-acclerator-known-issues"></a> Known Issues
+In version 1.5 of the Application Accelerator extension for VS Code, if an accelerator with [`custom types`](creating-accelerators/custom-types.hbs.md) is used _and_ the custom type form contains checkboxes, reprioritizing the custom types entries will cause the data to not properly reorder. This is a known bug and will be resolved in an upcoming release.
```
matt-dres, line 64:
```suggestion
In version 0.1.5 of the Application Accelerator extension for VS Code, if an accelerator with [`custom types`](creating-accelerators/custom-types.hbs.md) is used _and_ the custom type form contains checkboxes, reprioritizing the custom types entries will cause the data to not properly reorder. This is a known bug and will be resolved in an upcoming release.
```

partials/application-accelerator/_configuration.hbs.md
matt-dres: Will revert back to `-`

PR #2069, opened 2023-01-05
SSO: Clarify instructions for securing workload

eleftherias:
The `client.yaml` file in the sample application has the following format:
```
#@ load("@ytt:data", "data")

---
apiVersion: sso.apps.tanzu.vmware.com/v1alpha1
kind: ClientRegistration
metadata:
  name: #@ data.values.workload_name
  namespace: #@ data.values.namespace
spec:
  authServerSelector:
    matchLabels:
      name: #@ data.values.auth_server_name
# ...
```

This means that the client registration will try to match the authorization server based on the label called "name" and not the name of the resource.

This is especially confusing because in the documentation for [Provision an AuthServer
](https://docs.vmware.com/en/VMware-Tanzu-Application-Platform/1.3/tap/GUID-app-sso-getting-started-provision-auth-server.html) the resource name (i.e my-authserver-example) and the label "name" (i.e my-first-auth-server) are different.

1.3.x, 1.4.x

### DIFF
--- a/app-sso/app-operators/tutorials/securing-first-workload.hbs.md
+++ b/app-sso/app-operators/tutorials/securing-first-workload.hbs.md
@@ -131,7 +131,7 @@ as described.
 
 Apply the `client.yaml` definition file (described [above](#the-sample-applications-clientregistration))
 
-> **Caution** Make sure to set `auth_server_name` field to the name of the AuthServer custom resource.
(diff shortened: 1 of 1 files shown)


eleftherias:
Hey @ddubson could you or someone from you team confirm that this change makes sense?

ddubson (approved):
This looks great. 

I'm noticing a few other inconsistencies on this guide that AppSSO team will take care of to make this document a bit more readable.

ddubson:
@eleftherias can you also apply this TAP 1.4.x?

PR #2084, opened 2023-01-06
Adding in release notes & missing CLI doc

matt-dres:

1.4

### DIFF
--- a/cli-plugins/accelerator/command-reference/tanzu_accelerator_generate_from_local.hbs.md
+++ b/cli-plugins/accelerator/command-reference/tanzu_accelerator_generate_from_local.hbs.md
@@ -1,0 +1,50 @@
+## `tanzu accelerator generate-from-local`
+
+Generate a project from local or registered accelerators/fragments
+
+## Synopsis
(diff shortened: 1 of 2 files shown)


cli-plugins/accelerator/command-reference/tanzu_accelerator_generate_from_local.hbs.md
```diff
@@ -0,0 +1,50 @@
+## `tanzu accelerator generate-from-local`
```
matt-dres, line 1: @eleftherias - please review and let me know if anything should be revised.

eleftherias: I believe we should also update tanzu_accelerator.hbs.md with a reference to the generate-from-local docs https://github.com/pivotal/acc-tanzu-cli/blob/main/docs/tanzu_accelerator.md

matt-dres: ahhh good point, I'll update that. Thanks!

PR #2109, opened 2023-01-09
appsso storage service binding type now optional

ddubson:
Which other branches do you want a technical writer to cherry-pick this PR to (if any)? TAP 1.5 only.


YugangW:
Hi @ddubson, Just a quick heads-up that we don't have a branch for TAP 1.5 yet. The current main branch is intended for TAP 1.4 only. If this PR is for 1.5 **ONLY**, I would recommend resending the PR against `main` tomorrow afternoon when the new `main` branch is ready... Thank you and sorry for any inconvenience caused...

PR #2116, opened 2023-01-09
Adds configuring public clients docs for AppSSO ClientRegistration CRD

ddubson:
Which other branches do you want a technical writer to cherry-pick this PR to (if any)? TAP 1.5


YugangW:
Hi @ddubson, Does this PR apply to both 1.4 and 1.5 or 1.5 only? Thanks!

ddubson:
@YugangW - we made the decision on the team that it should apply only to 1.5. Thanks!

PR #2128, opened 2023-01-13
add note re: "_ytt_lib" in path

heyjcollins:


heyjcollins:
going to create another based on the latest docs code.

PR #2132, opened 2023-01-16
Include custom types as option for dataType

eleftherias:
1.5 and 1.4.1


RichardJJG:
@eleftherias, this applies to TAP v1.4.0 as well, I'm guessing? The `main` branch is now used for v1.5 content.

eleftherias:
@RichardJJG Yes, this applies to 1.4.0 as well

PR #2133, opened 2023-01-16
draft: 1 4 0 arr trust issues

Petahhh:
* This can be completely cherry-picked to `1-4-1`

This should also be cherry-picked onto `main` but not `d0c6082ca67ce9009e1fad0c990a76b54c57102a` because hopefully we fix it before 1.5

### DIFF
--- a/troubleshooting-tap/troubleshoot-using-tap.hbs.md
+++ b/troubleshooting-tap/troubleshoot-using-tap.hbs.md
@@ -594,3 +594,81 @@ take "/" at the end of the string.
 ### Solution
 
 Do not append "/" to the end of the string.
+
+---
+
+## <a id='cluster-issuer-trust-issues'></a> x509: certificate signed by unknown authority
+
+### Explanation
+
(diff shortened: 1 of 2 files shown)


troubleshooting-tap/troubleshoot-using-tap.hbs.md
```diff
@@ -594,3 +594,81 @@ take "/" at the end of the string.
...
+
+1. Extract the ClusterIssuer's Certificate Authority from cert-manager
+
+  ```console
+  kubectl get secret tap-ingress-selfsigned-root-ca -n cert-manager -o yaml | yq .data | cut -d' ' -f2 | head -1 | base64 -d
+  ```
+
+1. Add the certificate to the list of trusted certificate authorities. This is done by updating your tap-values.yml to append the above certificate authority to the field `shared.ca_cert_data`
+
+1. Pause the meta package's reconciliation. This prevents Tanzu Application Platform from reverting to the original values.
```
jrshaffe, line 629: I don't think you want to turn off the reconciliation.  You have to if you are trying to reinstall a specific component and don't want tap to reset it back to the original values,  but in this case you want the meta package to pickup that change and propagate it down to all the packages.

HenryBorys:
@jrshaffe I reverted the merge. Sorry, I had not noticed this was in draft state.

HenryBorys:
@Petahhh Is this ready to merge or still in draft state?

PR #2146, opened 2023-01-19
[TSCC-145] Remove outdated warning

pivotal-todd-ritchie:
We now support multiple pipelines per namespace.


HenryBorys:
@pivotal-todd-ritchie Is this just for 1.5 or does it apply to any other release? I need to know. Thanks.

pivotal-todd-ritchie:
It applies to 1.3 and 1.4 as well.

PR #2163, opened 2023-01-26
Added clarification based on customer feedback for JIRA TANZUSC-2844

krisapplegate:
Applies to all versions of TAP


vrabbi:
Why is iterate not added here as well?

RichardJJG:
I've merged this PR into main and edited it. If substantial further changes are needed, there will need to be a new PR.

RichardJJG:
I've received confirmation from Kris that this PR, in fact, only applies to the TAP v1.2 docs and later.

PR #2175, opened 2023-01-31
Add section for workload internal communication

izabelacg:
1-4-0 and 1-4-1

### DIFF
--- a/workloads/web.hbs.md
+++ b/workloads/web.hbs.md
@@ -57,3 +57,38 @@ tanzu apps workload apply tanzu-java-web-app --type=web
 
 You can use this to test which applications can function well as serverless web applications,
 and which are more suited to the `server` application style.
+
+## <a id="using"></a> Communication between `web` workloads
+
+When a workload of the type `web` is created, a Knative service is deployed to the cluster. To access your application,
+you will need the URL for the route created by the Knative Service. You can obtain it by running one of the commands below:
+
+```console
+tanzu apps workload get <WORKLOAD-NAME> --namespace <YOUR-DEVELOPER-NAMESPACE>
+kubectl get ksvc <WORKLOAD-NAME> -n <YOUR-DEVELOPER-NAMESPACE> -ojsonpath="{status.url}"
+```
+
+It is important to note that for communication between workloads within the same Kubernetes namespace, workloads of the type
+`web` should be addressable by referencing both the workload name and the workload's namespace at a minimal. This behaviour is
+currently distinct from workloads of the type `server`, which do not rely on the namespace name to establish service to service
+communication between applications within the same namespace.
+
+Follow the example below to understand the difference between service to service communication for `web` and `server`
+workloads. Let's suppose we have 3 applications deployed to the namespace called `dev-namespace`:
+1. one is a workload of the type `server` named `server-workload`
+2. the other is a workload of the type `web` named `web-workload`
+3. the third one is a pod running the busybox image with curl, named `busybox`
+
+If we get a shell to the running container of the `busybox` pod and send a request to the `server` workload using curl, the
+command would look like this:
+
+```console
+kubectl exec busybox -n dev-namespace -- curl server-workload.svc.cluster.local -v
+```
+
+On the other hand, if we try to reach the `web` workload from the `busybox` pod, the command will look slightly
+different since we will need to specify the namespace:
+
+```console
+kubectl exec busybox -n dev-namespace -- curl web-workload.dev-namespace.svc.cluster.local -v
+```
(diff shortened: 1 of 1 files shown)


workloads/web.hbs.md
```diff
@@ -57,3 +57,38 @@ tanzu apps workload apply tanzu-java-web-app --type=web
 
 You can use this to test which applications can function well as serverless web applications,
 and which are more suited to the `server` application style.
+
+## <a id="using"></a> Communication between `web` workloads
```
jsanin-vmw, line 61: You will need a unique anchor id. The id "using" is already taken by other title

jsanin-vmw (changes requested):

izabelacg:
Pls let me know if the content meets expectation for [PDY-1431](https://jira.eng.vmware.com/browse/PDY-1431). cc @tshaine @evankanderson

workloads/web.hbs.md
```diff
@@ -57,3 +57,38 @@ tanzu apps workload apply tanzu-java-web-app --type=web
 
 You can use this to test which applications can function well as serverless web applications,
 and which are more suited to the `server` application style.
+
+## <a id="communication"></a> Communication between `web` workloads
```
evankanderson, line 61: I think this might better be called "Calling `web` workloads within a cluster" -- this applies even if a Deployment from a `worker` workload wants to call a `web` workload.

workloads/web.hbs.md
```diff
@@ -57,3 +57,38 @@ tanzu apps workload apply tanzu-java-web-app --type=web
...
+When a workload of the type `web` is created, a Knative service is deployed to the cluster. To access your application,
+you will need the URL for the route created by the Knative Service. You can obtain it by running one of the commands below:
+
+```console
+tanzu apps workload get <WORKLOAD-NAME> --namespace <YOUR-DEVELOPER-NAMESPACE>
+kubectl get ksvc <WORKLOAD-NAME> -n <YOUR-DEVELOPER-NAMESPACE> -ojsonpath="{status.url}"
+```
+
+It is important to note that for communication between workloads within the same Kubernetes namespace, workloads of the type
+`web` should be addressable by referencing both the workload name and the workload's namespace at a minimal. This behaviour is
```
evankanderson, line 72: I think you mean "minimum" here, but this seems a little indirect.  Maybe "when calling a Knative service, the Service name and namespace are required"?

workloads/web.hbs.md
```diff
@@ -57,3 +57,38 @@ tanzu apps workload apply tanzu-java-web-app --type=web
...
+```console
+kubectl exec busybox -n dev-namespace -- curl server-workload.svc.cluster.local -v
+```
+
+On the other hand, if we try to reach the `web` workload from the `busybox` pod, the command will look slightly
+different since we will need to specify the namespace:
+
+```console
+kubectl exec busybox -n dev-namespace -- curl web-workload.dev-namespace.svc.cluster.local -v
+```
```
evankanderson, line 94: This format works for both web and server workloads, right?

Should we recommend using this format by default?

workloads/web.hbs.md
```diff
@@ -57,3 +57,38 @@ tanzu apps workload apply tanzu-java-web-app --type=web
...
 and which are more suited to the `server` application style.
+
+## <a id="communication"></a> Communication between `web` workloads
+
+When a workload of the type `web` is created, a Knative service is deployed to the cluster. To access your application,
+you will need the URL for the route created by the Knative Service. You can obtain it by running one of the commands below:
+
+```console
+tanzu apps workload get <WORKLOAD-NAME> --namespace <YOUR-DEVELOPER-NAMESPACE>
+kubectl get ksvc <WORKLOAD-NAME> -n <YOUR-DEVELOPER-NAMESPACE> -ojsonpath="{status.url}"
```
evankanderson, line 68: Do you want to use `status.address.url` for within-cluster URL? `status.url` will be internet-reachable while `status.address.url` is always internal.

workloads/web.hbs.md
izabelacg: Yes, works for both. In that section, I was trying to explain the different behaviour between web and server. Do you think I should remove that example then, and only recommend to using service name + namespace always

workloads/web.hbs.md
izabelacg: I will rephrase that, thanks!

workloads/web.hbs.md
izabelacg: I created another PR to address your comments, since this one was merged already: https://github.com/pivotal/docs-tap/pull/2177

PR #2185, opened 2023-02-09
fix: correct MavenArtifact example

garethjevans:

1.3


rashedkvm (approved):

HenryBorys:
@garethjevans I see you targeted this change for 1.5 (main) and requested it also be cherry picked to 1.3. It would seem it must also go to 1.4+ as well. So is it that you actually want it cherry picked to 1.3+ all the way through 1.5?

PR #2193, opened 2023-02-14
SCC Reference Update

waciumawanjohi:
Which other branches do you want a technical writer to cherry-pick this PR to (if any)? None, I will make a PR against 1.3 with a branch with small changes.


YugangW:
Hi @waciumawanjohi, Is this PR applicable to 1.4 docs? Thanks.

waciumawanjohi:
@YugangW I assumed that these changes would be applied to 1.4 and the coming 1.5 branch. Was my assumption correct? If not, is there anything I need to do (beyond this message) to make that the case?

PR #2200, opened 2023-02-14
Add more details for api-portal and api-auto-reg

shrutiyer:
These changes are only for TAP 1.5

### DIFF
--- a/api-auto-registration/installation.hbs.md
+++ b/api-auto-registration/installation.hbs.md
@@ -11,18 +11,14 @@ See [Tanzu Application Platform Prerequisites](../prerequisites.md).
 
 ## <a id='prereqs'></a> Using with TLS
 
-Starting in Tanzu Application Platform v1.4, TLS is turned on by default for
-several components. This causes an issue with API Auto Registration. For
-information about a workaround, see
-[Troubleshooting](../api-auto-registration/troubleshooting.hbs.md). 
+Starting in Tanzu Application Platform v1.4, TLS is turned on by default for several components.
+API Auto Registration automatically trusts the CA for the shared `ingress_issuer` when using the default ClusterIssuer `tap-ingress-selfsigned`.
+This change means that a `Certificate` is automatically generated using this issuer.
 
-Starting in Tanzu Application Platform v1.4.1, API Auto Registration
-automatically trusts the CA for the shared `ingress_issuer`. This change means
-that a `Certificate` is generated using this issuer. If you do not want a
-`Certificate` to generate automatically, you can set the `auto_generate_cert`
-flag to `false` in the values file. You must follow the instructions in the
-troubleshooting guide to manually set the `ca_cert_data`. See
-[Troubleshooting](../api-auto-registration/troubleshooting.hbs.md). 
+If you do not want a `Certificate` to generate automatically, you can set the `auto_generate_cert` flag to `false` in the values file.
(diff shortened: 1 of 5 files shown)


shrutiyer:
@karinepires @jrshaffe @mikegehard any edits?

api-auto-registration/installation.hbs.md
```diff
@@ -11,18 +11,14 @@ See [Tanzu Application Platform Prerequisites](../prerequisites.md).
...
-troubleshooting guide to manually set the `ca_cert_data`. See
-[Troubleshooting](../api-auto-registration/troubleshooting.hbs.md). 
+Starting in Tanzu Application Platform v1.4, TLS is turned on by default for several components.
+API Auto Registration automatically trusts the CA for the shared `ingress_issuer` when using the default ClusterIssuer `tap-ingress-selfsigned`.
+This change means that a `Certificate` is automatically generated using this issuer.
+
+If you do not want a `Certificate` to generate automatically, you can set the `auto_generate_cert` flag to `false` in the values file.
+If you want to replace the default with a custom ingress issuer, see [Security and compliance](../security-and-compliance/about.hbs.md).
+For both these scenarios, whenever not using the default ClusterIssuer `tap-ingress-selfsigned` or not automatically generating certificates
+or using other custom CAs, you will need to manually generate and set the cert. See [Troubleshooting](../api-auto-registration/troubleshooting.hbs.md#set-ca-crt).
```
jrshaffe, line 21: I think we can remove "generate", they will need to locate the CA PEM, but they shouldn't need to generate any certificates (all though that is an option).  We only have to generate one because we need to get the CA into a secret located in our namespace to mount to the pod, but the same CA can be extracted from an existing secret in the cert-manager namespace, and for some of the ClusterIssuer's generating a Certificate won't result in a secret that contains the CA PEM, for instance if they are using Vault the ca won't be in the secret that get's generated, they have to go and get the CA from vault itself.

jrshaffe (approved):

api-auto-registration/installation.hbs.md
shrutiyer: See the latest for the updated wording. Do you think we should add more detail for this?

karinepires (approved):
Great work Shruti!
I really enjoyed the hints with "VMware recommends / do not recommend" as a way to inform the users for better decisions!

api-portal/install-api-portal.hbs.md
```diff
@@ -70,3 +75,57 @@ To install API portal:
...
+    kubectl get pods -n api-portal
+    ```
+
+## <a id='update-values'></a>Update the install values for api-portal package
+
+1. To overwrite the default values, create new values, or update the existing values, you will need an
+   `api-portal-values.yaml` file. If you do not already have an existing values file, you can extract the existing values using the command:
+
+    ```console
+    tanzu package installed get api-portal -n api-portal -f api-portal-values.yaml
```
visomar, line 97: If we follow the same steps mentioned before this section, the namespace used for installing api-portal should be `tap-install`:
```suggestion
    tanzu package installed get api-portal -n tap-install -f api-portal-values.yaml
```

visomar (approved):
This is perfect, I love how well expressed it is.
I might even take a bit of it to be used in API portal's documentation

image

I found a tiny thing in one of the examples

PR #2218, opened 2023-03-01
Tap gui/1.5.0 bug bash docs

mstergianis:
This is only applicable to 1.5, so any branches that have been cut for that. I'm pretty sure that's just main though right?


RichardJJG:
Thanks, @mstergianis, and, yes, it's just `main`.

PR #2219, opened 2023-03-02
Update install-airgap file to correct the indentation for shared.ca_c…

dhadukk:
…ert_data and remove policy controller package exclude.


dhadukk:
Please add this change for TAP 1.3 and TAP 1.4

PR #2220, opened 2023-03-02
Merge branch 'main' of https://github.com/pivotal/docs-tap

andreizimin:
# Which other branches do you want a technical writer to cherry-pick this PR to (if any)? 1.5

This doc needs to have the url https://docs.vmware.com/en/VMware-Tanzu-Application-Platform/1.5/tap/tap-portal-telemetry.html


andreizimin:
the email in the text should be tap-pendo@groups.vmware.com and NOT @vmware.com

PR #2230, opened 2023-03-06
Remove unnecessary commands to get all tags from repository

joaopapereira:
This PR should be merged when Cluster Essentials 1.5 is out


RichardJJG:
It's safe to merge this into `main` right now because `main` is for TAP v1.5 and has only staged docs. In the unlikely event that Cluster Essentials v1.5 is released after TAP v1.5, a writer (most likely @YugangW) can hide this content until Cluster Essentials v1.5 is released.

PR #2232, opened 2023-03-06
Add screenshots of telemetry reports

mgibson1121:

1.3, 1.4, main


mgibson1121:
In the second commit, I hard coded a reference to a location in TAP install docs [Full profile install section](https://docs.vmware.com/en/VMware-Tanzu-Application-Platform/1.4/tap/install.html#full-profile
). Could a docs writer update to use the correct format?  I believe what I've done her is likely incorrect.

PR #2245, opened 2023-03-10
Reference Backstage Identity

andreizimin:
1.4 and 1.5 please


tap-gui/images/backstage-identity.png
doddatpivotal, line 1: Can use a screenshot where the user belongs to at least one group so that owernship entity looks different from user entity?

andreizimin: @waldirmontoya25 , @jamieklassen do you have an instance/example as per Dodd's requirement?

waldirmontoya25: @andreizimin I don't have a TAP Portal demo instance running :(

jamieklassen: I cooked one up for you:
ownership_entities

andreizimin: Thank you @jamieklassen ! I think this looks great!

PR #2268, opened 2023-03-17
Document App Accelerator Provenance mechanism

ericbottard:
Fix https://github.com/pivotal/acc-release/issues/502

None. This is a TAP 1.5 feature


kcoriordan:
tracked in https://jira.eng.vmware.com/browse/TAAP-1129

PR #2273, opened 2023-03-19
updated tap-auth docs to use latest pinniped version

dherbrich:
# Which other branches do you want a technical writer to cherry-pick this PR to (if any)? back to all still supported TAP versions


YugangW:
Hi @dherbrich, Looks like there is conflict, would you mind helping resolve so I can get it merged? Thanks!

dherbrich:
i think this way it is up to date and also works if we merge all changes back to the last support TAP version. the doc now says to install pinniped 0.22 and appropriate CLI.

PR #2323, opened 2023-03-30
prisma - add section to describe how to configure a proxy, update rego for prisma defined scan policy

rabeyta:
1-4-2


Mpluya:
LGTM

PR #2330, opened 2023-03-30
Fixed bookmark links on Tanzu CLI install page

ncarlson:

* 1-4-1
* 1-4-0

### DIFF
--- a/install-tanzu-cli.hbs.md
+++ b/install-tanzu-cli.hbs.md
@@ -2,17 +2,17 @@
 
 This topic describes how to:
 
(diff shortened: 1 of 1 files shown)


kcoriordan:
Reverting as this breaks a lot of bookmarks/anchors in other files. The IDs for these headings are used in links in other topics.
The table of contents at the top of this file is not required. We have a right navigation which has all of these headings. So the table of contents is duplicate content and not required. I will rewrite the short description so it includes key terms for SEO.

Rather than fixing all the other files and cherry-picking to 1-4-4, I am reverting, and I will make a commit directly to 1-4-4

PR #2343, opened 2023-04-03
allow client namespaces annotation is optional in AuthServer

ddubson:
# Which other branches do you want a technical writer to cherry-pick this PR to (if any)? TAP 1.6 only


YugangW:
Hi @ddubson, Since this PR targets TAP 1.6, I will not merge it till after the TAP 1.5 release.

PR #2346, opened 2023-04-03
add delivery with carvel app page

nader-ziada:

none


scc/delivery-with-carvel-app.hbs.md
```diff
@@ -0,0 +1,210 @@
...
+
+      deploy:
+     - kapp:
+         intoNs: <DESIRED_NAMESPACE>
+         rawOptions: ["--dangerous-allow-empty-list-of-resources=true"]
+   ```
+
+   > **Note:** The fetch section can includes entries for all the locations in the gitops repo to be deployed, and be appended with other run clusters if needed. Each App CR must specify either a service account (via `spec.serviceAccountName`) or, a `Secret` with kubeconfig contents for some Run cluster (via `spec.cluster.kubeconfigSecretRef.name`).
+
+3. You will also need to create a `ServiceAccount` with associated RBAC permissions for the App to use. Example of the required permissions for the `ServiceAccount`. Make sure to assess appropriate RBAC needed for your specific PackageInstalls.
```
nickjameswebb, line 153: I think we need more clarification on service accounts. We have a section up above where we create a service account on a run cluster, and then this section for the build cluster. I don't think it's clear where the user should be using both of those service accounts. Also, this section should be before we create the `App` because it's a prereq.

scc/delivery-with-carvel-app.hbs.md
```diff
@@ -0,0 +1,210 @@
...
+   apiVersion: rbac.authorization.k8s.io/v1
+   metadata:
+     name: pkg-gitops-pkgi-role
+     namespace: <DESIRED_NAMESPACE>
+     annotations:
+       kapp.k14s.io/change-group: "packageinstall-setup"
+   rules:
+   - apiGroups: ["*"]
+     resources: ["*"]
+     verbs: ["*"]
```
nickjameswebb, line 175: Need to replace with specific permissions

scc/delivery-with-carvel-app.hbs.md
```diff
@@ -0,0 +1,210 @@
...
+   ---
+   apiVersion: kappctrl.k14s.io/v1alpha1
+   kind: App
+   metadata:
+     name: hello-app-app
+     namespace: <build-cluster-ns>
+   spec:
+     # specifies that app should be deployed authenticated via the
+     # given service account, found in this namespace
+     serviceAccountName: <SERVICE_ACCOUNT>
```
nickjameswebb, line 121: Is this service account on the run cluster or  build cluster? Can we set both separately?

nader-ziada: Its specified in the note right after: Each App CR must specify either a service account (via `spec.serviceAccountName`) or, a `Secret` with kubeconfig contents for some Run cluster (via `spec.cluster.kubeconfigSecretRef.name`).

PR #2359, opened 2023-04-04
Update install-air-gap.hbs.md

dhadukk:
Remove username password from the value file and provide secret


mdc81:
@dhadukk steps to create the secret also should be included or referenced from TBS docs.

mdc81:
looks good

snneji:
Hi @dhadukk, is this just for 1.5 or should it be cherry picked to another version?

dhadukk:
This change is only required for TAP 1.5

PR #2378, opened 2023-04-04
IntelliJ: Using the plugin update

RuchikaGoel:
None


RichardJJG:
I can see that the text changes actually do apply to varying extents in 1-4-2, 1-3-6, and 1-2-2, and I've cherry-picked accordingly.

PR #2402, opened 2023-04-06
Update environment variables used in secrets

weresch:
None, just for TAP 1.5.0


weresch:
Updated environment variables as they can't use hyphens. Kept the styling consistent between the two pages re: use of {}

PR #2408, opened 2023-04-07
Add application accelerator IntelliJ docs for 1.5

nierajsingh:


DiegoWong:
The entry in the index file is missing, look at the file `toc.md` an add a reference to the docs there

DiegoWong (approved):

PR #2417, opened 2023-04-10
[VSEE-1012] Fix table formatting using HTML instead of github markdown

rxl7906:
1.5


snneji:
Hi @rxl7906, which TAP versions does this apply to? Does it apply to TAP 1.5 as well as 1.6 (main is now 1.6)?

PR #2419, opened 2023-04-10
Reorganize topics for Spring Cloud applications

emalm:
* Create separate topics for Spring Cloud Gateway, Spring Cloud Config, and Spring Cloud DiscoveryClient applications
* Organize topics under Getting Started instead of Integrations
* Update TOC and other index topics
* Incorporate changes from https://github.com/pivotal/docs-tap/commit/5c1a729a1220630f970848ce0a8425a46507c733

These changes are for the upcoming TAP 1.5 release.


emalm:
cc @julian-hj @csterwa @habuma @maly7

csterwa (approved):
I like the changes.

csterwa:
> * Create separate topics for Spring Cloud Gateway, Spring Cloud Config, and Spring Cloud DiscoveryClient applications
> * Organize topics under Getting Started instead of Integrations
> * Update TOC and other index topics
> * Incorporate changes from [5c1a729](https://github.com/pivotal/docs-tap/commit/5c1a729a1220630f970848ce0a8425a46507c733)
> 
> These changes are for the upcoming TAP 1.5 release.

Great changes! I think this is much more clear now with each on their own page.

As for cherry picking, I don't think that this is useful for pre-1.5 releases so I think it is only needed on `main` and `v1.5`, correct?

emalm:
Yep, it looks like the 1.5 docs are still building from the tip of `main`, but presumably a new `1-5` branch will show up for them this week. Since ACS is new in TAP 1.5, the Config and DiscoveryClient topics wouldn't apply to earlier TAP versions.

SCG for K8s is less coupled to the TAP release, though: would we support it being used with TAP 1.4 or earlier? If so, we could cherry-pick those topics onto docs for those earlier TAP versions, although I would see it as a nice to have.

csterwa:
SCG for k8s is only part of TAP sku as of 1.5 so if rather just leave it in 1.5 docs. There is sufficient info in SCG for k8s docs to support previous versions.

PR #2426, opened 2023-04-11
Provide alternative options for internal communication 

voor:
Provide alternative options for internal communication  and rename sections since alternatives are not all TLS based.


mzbat (changes requested):
Communication spelled incorrectly on line 138.

ryan-a-baker:
Looks good to me!  Thanks a ton for the suggestion @voor!

PR #2433, opened 2023-04-13
Enhance the details for existing apps

shrutiyer:
1.5.1 and 1.5.0 and 1.4.4


shrutiyer:
@csterwa for the Spring App Lift and Shift, I am thinking we should enhance the actual docs itself and try to link the relevant sections into the Spring docs. What do you think?

shrutiyer:
No rush. Any thoughts on this PR's changes? I am thinking for TAS and OSS migrations, where they have apps that expose an API, we can add a link to the "Using an existing Spring Boot project using springdoc" section where they will have all the details needed.

csterwa:
Yeah, I'm going to be updating the TAP docs section with a PR on SCG for K8s to point towards specific docs that are pertinent.

I think we can do the same in this case for API Auto Reg.

csterwa (approved):

RichardJJG:
@shrutiyer, the assigned writer will add it as part of the editing work.

PR #2493, opened 2023-05-09
Specify amr deploy bug in 1.5.1 release notes

julia-pu:
In TAP 1.5.1 there is a bug with the scst store, AMR deploy does not work. 

I was told to make a doc JIRA and update the release notes with this issue. I've made this [JIRA ticket](https://jira.eng.vmware.com/browse/TAAP-1432), please let me know if this is correct or if it's another JIRA I should be making


snneji:
Thanks @julia-pu, we've received your ticket and will update the docs as soon as we can.

PR #2494, opened 2023-05-10
add troubleshooting docs on defining imagePullSecrets

katmutua:
Add troubleshooting information for defining `imagePullSecrets`


snneji:
Hi @katmutua, is this only for TAP 1.6 or should it be in other versions as well?

katmutua:
This addition to the troubleshooting section is relevant to all versions.

PR #2502, opened 2023-05-12
Removes instructions to set service_type to ClusterIP for TAP GUI

mstergianis:
That's been the default since TAP 1.0

Any of them

### DIFF
--- a/install-offline/profile.hbs.md
+++ b/install-offline/profile.hbs.md
@@ -1,6 +1,6 @@
 # Install Tanzu Application Platform in your air-gapped environment
 
-This topic tells you how to install Tanzu Application Platform (commonly known as TAP) 
+This topic tells you how to install Tanzu Application Platform (commonly known as TAP)
 on your Kubernetes cluster and registry that are air-gapped from external traffic.
 
 Before installing the packages, ensure that you have completed the following tasks:
@@ -304,7 +304,6 @@ appliveview_connector:
       -----END CERTIFICATE-----
 
 tap_gui:
-  service_type: ClusterIP
   app_config:
     kubernetes:
       serviceLocatorMethod:
--- a/install-online/profile.hbs.md
+++ b/install-online/profile.hbs.md
@@ -1,6 +1,6 @@
 # Install Tanzu Application Platform package and profiles
 
-This topic tells you how to install Tanzu Application Platform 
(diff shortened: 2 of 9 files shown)


install-online/profile.hbs.md
```diff
@@ -297,8 +297,7 @@ buildservice:
     namespace: "KP-DEFAULT-REPO-SECRET-NAMESPACE"
 
 tap_gui:
-  service_type: ClusterIP # If the shared.ingress_domain is set as earlier, this must be set to ClusterIP.
-  metadataStoreAutoconfiguration: true # Create a service account, the Kubernetes control plane token and the requisite app_config block to enable communications between Tanzu Application Platform GUI and SCST - Store.
+  metadataStoreAutoconfiguration: true # Creates a service account, the Kubernetes control plane token and the requisite app_config block to enable communications between Tanzu Application Platform GUI and SCST - Store.
```
andreizimin, line 300: SCST - supply chain security tools (took me back to the TAP Docs to decipher that one)

mstergianis: I think it's the standard nomenclature for metadata store (which also has the name AMR apparently)

```shell
$ ag 'SCST - Store' | wc -l
117
```

mstergianis: Anyways the only reason that line is appearing in the diff at all is because I fixed a typo that fell through the cracks.

It used to be "Create a service account" and I made it "Creates a service account"

install-offline/profile.hbs.md
```diff
@@ -1,6 +1,6 @@
 # Install Tanzu Application Platform in your air-gapped environment
 
-This topic tells you how to install Tanzu Application Platform (commonly known as TAP) 
+This topic tells you how to install Tanzu Application Platform (commonly known as TAP)
```
andreizimin, line 3: Do we need to make changes in line 337 - also mentions ClusterIP

RichardJJG:
Thanks, I've cherry-picked as far back as the v1.3 docs. The v1.2 docs are so different that I can't confidently make those code changes (and with the v1.2 docs being archived soon, it doesn't seem worth the headache...but tell me if I'm wrong about that).

mstergianis:
Thanks @RichardJJG! Yeah no point if they're being archived soon

PR #2510, opened 2023-05-19
Flux Source Controller Release Note

rashedkvm:
n/a


YugangW:
Hi @rashedkvm, Thanks for sending in the PR. Regarding the deprecation section, have all the mentioned fields been removed or are they still available but deprecated in favor of the new fields?

rashedkvm:
> Regarding the deprecation section, have all the mentioned fields been removed or are they still available but deprecated in favor of the new fields?

Those fields are deprecated, but still available. Please note, it will be removed in the next release.

PR #2517, opened 2023-05-24
Fix RDS yaml formatting

Samze:

1.5 and beyond


Samze:
cc @snneji

PR #2527, opened 2023-05-30
STK, Crossplane, Bitnami updates for TAP 1.6

teddyking:
## About

Updates docs for Services Toolkit, Bitnami Services and Crossplane TAP components for TAP 1.6.

Which other branches do you want a technical writer to cherry-pick this PR to (if any)? N/A


crossplane/how-to-guides/use-pre-existing-crossplane-tap.hbs.md
```diff
@@ -0,0 +1,41 @@
+# Use a pre-existing or out-of-band installation of Crossplane with Tanzu Application Platform
+
+This topic describes options for installing Tanzu Application Platform to a cluster which already has an installation of Crossplane, or for which you wish to install Crossplane using an out-of-band mechanism (e.g. through Helm install).
+
+## About
+
+The crossplane.tanzu.vmware.com Package is included in Tanzu Application Platform's Full, Iterate and Run profiles. Any installation of Tanzu Appliation Platform using one of these profiles will result in the installation of Crossplane to the cluster via the Package. In most cases this is desirable behaviour as Crossplane is required to support many of the services capabilities offered by Tanzu Application Platform.
```
st3v, line 7:
```suggestion
The crossplane.tanzu.vmware.com Package is included in Tanzu Application Platform's Full, Iterate and Run profiles. Any installation of Tanzu Appliation Platform using one of these profiles will result in the installation of Crossplane to the cluster via the Package, unless that Package has been explicitly excluded. In most cases, this is desirable behaviour as Crossplane is required to support many of the services capabilities offered by Tanzu Application Platform.
```

crossplane/how-to-guides/use-pre-existing-crossplane-tap.hbs.md
```diff
@@ -0,0 +1,41 @@
...
+- crossplane.tanzu.vmware.com
+```
+
+Choose this option if you already have Crossplane installed on your cluster, you wish to continue to manage your Crossplane installation out-of-band, and you are either installing or upgrading to Tanzu Application Platform >= v1.5.0.
+
+This option has the effect of completely disabling the crossplane.tanzu.vmware.com Package from the Tanzu Application Platform install, which leaves you free to manage the Crossplane installation in whatever way you see fit.
+
+It is important to note that other Tanzu Application Platform components depend on the availability of Crossplane. One such example is [Bitnami Services](../../bitnami-services/about.hbs.md). Therefore if you are planning to exclude the crossplane.tanzu.vmware.com Package, it is recommended to first install Crossplane using whatever mechanism you choose prior to installing Tanzu Application Platform with the Package excluded.
+
+## Option 2: Adopt existing Crossplane resources
```
st3v, line 28: I wonder if we should actually document this option so prominently. There might be certain configurations of Crossplane with existing managed resources that might be changed during adoption in ways that we haven't foreseen and hence might cause issues for the user. We haven't tested this in-depth and I'm not even sure we can reliably do that considering that there are essentially unlimited configuration permutations for Crossplane.

st3v: At the very least we might want to add some sort of warning.

crossplane/install-crossplane.hbs.md
```diff
@@ -3,6 +3,8 @@
 This topic tells you how to install Crossplane from the Tanzu Application Platform package
 repository.
 
+> **Note** If installing Tanzu Application Platform to a cluster which already has a pre-existing installation of Crossplane, please refer to [Use a pre-existing or out-of-band installation of Crossplane with Tanzu Application Platform](how-to-guides/use-pre-existing-crossplane-tap.hbs.md)
```
st3v, line 6: Would we want to add a note on the `PackageInstall` reconciliation error for cases where the user missed this note?

st3v: I guess there's already a high-level note on the reconciliation error in the release notes. But that is a bit hidden and maybe not specific enough.

crossplane/how-to-guides/use-pre-existing-crossplane-tap.hbs.md
teddyking: Hmm fair point. I think we probably need _some_ documentation given it will show up in the list of package values and users can configure it ... perhaps I'll just add it as more of a small note to the end of the doc and also mention that milage may vary ...

crossplane/install-crossplane.hbs.md
teddyking: > Would we want to add a note on the PackageInstall reconciliation error

Not sure exactly what you mean? Do you mean to improve the error message displayed on the reconcile error?

st3v: Nope, just telling the reader what error to expect in that case...

The`PackageInstall` for the `crossplane.tanzu.vmware.com` package will fail to reconcile with the following error: `Resource already exists. Consider excluding package in tap-values`.

teddyking: Gotcha makes sense - I'll add that in.

crossplane/how-to-guides/use-pre-existing-crossplane-tap.hbs.md
teddyking: Actually I think it's better to remove mention of this option all together. Feels a bit weird to document "here is an option but we don't recommend it" ...

snneji:
Thanks @teddyking

PR #2548, opened 2023-06-05
Add info on authenticating to Azure Devops repos

waciumawanjohi:

This change should apply to TAP 1.7, 1.6, 1.5.

### DIFF
--- a/scc/azure.hbs.md
+++ b/scc/azure.hbs.md
@@ -16,6 +16,17 @@ For information about how Azure DevOps is different from other Git providers, se
 
 The operator requires special configuration to integrate Azure DevOps repositories into a supply chain.
 
+## <a id="Azure Auth"></a> Azure Authentication
+
+Documentation for configuring secrets to authenticate with your Azure Devops git repository can be found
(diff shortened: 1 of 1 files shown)


scc/azure.hbs.md
```diff
@@ -16,6 +16,17 @@ For information about how Azure DevOps is different from other Git providers, se
...
+## <a id="Azure Auth"></a> Azure Authentication
+
+Documentation for configuring secrets to authenticate with your Azure Devops git repository can be found
+[here](./git-auth.hbs.md). Note that Azure http/s auth requires:
+```yaml
+username: "_token"
+password: AZURE-USER-TOKEN
+```
+
+See Microsoft documentation for [Azure Devops Personal Access Tokens](https://learn.microsoft.com/en-us/azure/devops/organizations/accounts/use-personal-access-tokens-to-authenticate).
```
zrob, line 28: can we also make this note in the docs at https://docs.vmware.com/en/VMware-Tanzu-Application-Platform/1.5/tap/scc-git-auth.html

so users can see it all in one place

make a note about azure support.  also is there a similar thing for gitlab?

PR #2567, opened 2023-06-09
VSEE-1072 | Added sample IVS for Snyk

chaomonica:
- 1.6 only

### DIFF
--- a/scst-scan/custom-ivs-samples.md
+++ b/scst-scan/custom-ivs-samples.md
@@ -1,8 +1,30 @@
 # Configure custom ImageVulnerabilityScan samples
 
-This topic explains how to configure an ImageVulnerabilityScan sample for CarbonBlack scans with Supply Chain Security tools - Scan.
+This topic lists out sample ImageVulnerabilityScans utilizing various scanners and associated secrets. The following samples provide a scaffold for you to use with the below scanners:
+- Carbon Black
+- Snyk
 
-## Configure an ImageVulnerabilityScan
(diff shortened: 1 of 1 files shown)


chaomonica:
Hi @ryan-a-baker , this PR creates a new common general instructions section (about common fields the user needs to modify) as well as adds in the generic IVS for Snyk sample. Could you please review the changes?

ryan-a-baker:
This looks good to me.  I think one of the things I'm still trying to wrap my head around is how the IVS templates translate to ClusterImageTemplate.  I can see how we take these IVS templates and run a standalone scan, but that won't be a common use case.  We'll want to make sure that we have instructions on how to use these in conjuction with the cluster image template so that they can be used in the supply chains.  Now that I'm undersatnding this all a little bit better, I have to wonder if posting IVS templates is the right thing to do or would it be better for them to be a ClusterImageTemplate?  I'm not really sure.  As it stands right now I don't think a user could just pull this Snyk sample and use it without in a supply chain without the extra work of merging it with a ClusterImageTemplate?

In the context of this PR, this PR Is fine, but I think we can pull some organization together at a higher level.

chaomonica:
@ryan-a-baker Yes, a user would need to take this IVS Snyk sample and embed it into a clusterimagetemplate before it could be used in a supply chain. The last step of "General usage instructions" does direct the user to the [clusterimagetemplate page](https://github.com/pivotal/docs-tap/blob/main/scst-scan/clusterimagetemplates.hbs.md) but that page does need more instructions on directing the user to incorporate a premade IVS in.

PR #2581, opened 2023-06-12
VSEE-998 | Included activeKeychains and updated instructions in authoring clusterimagetemplate

chaomonica:
-1.6 only


chaomonica:
This MR addresses changes that will be introduced in the ootb MR 336 as well as improvements from previous feedback [here](https://github.com/pivotal/docs-tap/pull/2567#issuecomment-1584722071).

PR #2593, opened 2023-06-13
Clarify the use of CA certificates in Flux, ESO, and Tekton

rszumlakowski:

~~TAP 1.7~~
TAP 1.6


snneji:
Hi @rszumlakowski, main is currently at 1.6. If this PR is for 1.7, could you put it into draft mode until 1.6 is released so that we don't accidentally merge it into 1.6. Thanks

rszumlakowski:
Actually, can we merge it to TAP 1.6 instead?

PR #2604, opened 2023-06-15
TANZUSC-3691 Azure and aws tap installation inconsistencies

leclaist:
This PR attempts to solve some of the issues here: https://jira.eng.vmware.com/browse/TANZUSC-3691

* Addresses inconsistencies in registry IAM role creation 
* Includes image relocation steps 
* May address runtime error on registry secret creation 
* tap-values.yaml templated similarly to aws


RichardJJG:
@leclaist, which other branches do you want a technical writer to cherry-pick this PR to (if any)?

PR #2609, opened 2023-06-19
docs(tap-gui): don't say "external" ClusterIssuer

mamachanko:
The idea of an "external" (Cluster)Issuer is a specific concept of cert-manager. It refers to (Cluster)Issuer which are maintained out-of-tree. This makes it possible to extend cert-manager with implementations for almost any CA. (See:
https://cert-manager.io/docs/configuration/external/)

Hence, the phrase "external ClusterIssuer" is misleading as it is used by TAP GUI's TLS docs. Just saying "ClusterIssuer" is better and just as precise.

⚠️ This should be a approved by a member of TAP-GUI as this is an unsolicited, external 😄  contribution.

---

# Target

`1.6.x`

_

### DIFF
--- a/tap-gui/tls/cert-manager-clusterissuer.hbs.md
+++ b/tap-gui/tls/cert-manager-clusterissuer.hbs.md
@@ -1,4 +1,4 @@
-# Configure a TLS certificate by using cert-manager and an external ClusterIssuer
+# Configure a TLS certificate by using cert-manager and a ClusterIssuer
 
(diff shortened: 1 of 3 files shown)


RichardJJG:
Thanks, @mamachanko, I'll get approval from a TAP GUI SME before merging

mstergianis:
LGTM. Thanks for watchin our backs :smile:

PR #2610, opened 2023-06-19
Update naming of data services

teddyking:
## About

This PR updates naming from "VMware Tanzu RabbitMQ" to "VMware RabbitMQ for Kubernetes" and "VMware Tanzu Postgresql" to "VMware SQL with Postgres for Kubernetes"

Which other branches do you want a technical writer to cherry-pick this PR to (if any)? N/A


RichardJJG:
Should a writer cherry-pick this back to earlier versions, @teddyking?

teddyking:
Actually yes it probably makes sense to cherry-pick this back to TAP 1.5 as well, thanks!

PR #2630, opened 2023-06-22
CNR - Update docs for version 2.3.1

jsanin-vmw:
No need to cherry-pick

This is for tap 1.6.0


xtreme-vikram-yadav:
lgtm

PR #2637, opened 2023-06-22
Fix path to vault scripts in gitops-ri 0.2.0

neil-hickey:


RichardJJG:
@neil-hickey, which other branches do you want a technical writer to cherry-pick this PR to (if any)?

PR #2651, opened 2023-06-26
[TAP 1.6] tbs/deps: Update buildpack, stack versions

arjun024:
- Also remove Bionic based stacks that are not shipped with TAP 1.6
- Also remove Procfile from lite deps, as I believe it was a docs oversight. Procfile doesn't seem to have ever existed in lite.

### Which other branches do you want a technical writer to cherry-pick this PR to (if any)?
This change only applies to TAP 1.6


arjun024:
@johnnyr0x

PR #2675, opened 2023-06-30
Add overlay for source scanning

ryan-a-baker:


snneji:
Hi @ryan-a-baker, is this only for TAP 1.6 or should we cherry pick it to other versions?

PR #2749, opened 2023-07-13
Migrate the spec.location to the new format and announce deprecation

LittleBaiBai:
This is for TAP 1.7 only.


shrutiyer (approved):
Just have to remember to also add this to release notes here

PR #2750, opened 2023-07-14
Updates AWS Resources Page for TAP 1.6

doddatpivotal:
Includes additional capabilities included in TAP 1.6 which where not present in previous releases.

This is for TAP 1.6


ryan-a-baker:
Thank you @doddatpivotal!!!

PR #2794, opened 2023-07-27
[Gitops] Add error case for secret engine of vault

neil-hickey:
# https://jira.eng.vmware.com/browse/TANZUSC-4314


neil-hickey:
Hi folks! I requested merge to `main`. I think that's still correct? I want this to be part of the 1.6.x docs and onwards

PR #2802, opened 2023-08-01
appsso: add token viewer doc to verify token

sarahwalther:

- 1.5.3
- 1.4.8
- 1.4.7


app-sso/how-to-guides/service-operators/verify-token-settings.hbs.md
```diff
@@ -0,0 +1,192 @@
+# Verify token settings
```
mamachanko, line 1: **[thought, non-blocking]:** In the TOC there are sections "token settings", "token signatures" and then "verify token settings". It seems as if the last relates to the first, so why does it come last? Could it be merged into "token settings" for cohesion?

app-sso/how-to-guides/service-operators/verify-token-settings.hbs.md
```diff
@@ -0,0 +1,192 @@
...
+  selector:
+    matchLabels:
+      name: test-application
+  template:
+    metadata:
+      labels:
+        name: test-application
+    spec:
+      containers:
+        - image: bitnami/oauth2-proxy:7.3.0
```
mamachanko, line 69: **[thought, non-blocking]:** I wonder if we should include a note/warning that this is pulling in an unvetted, public image even though its bitnami. Also, it won't work behind an air-gap.

mamachanko (approved):

app-sso/how-to-guides/service-operators/verify-token-settings.hbs.md
sarahwalther: I went back and forth on this one and didn't know which I preferred and ended up just making a decision to move forward. I'm happy to move it!

app-sso/how-to-guides/service-operators/verify-token-settings.hbs.md
sarahwalther: Great call! Yes, we definitely should include a warning. And yes, I will also add the air gap note.

YugangW:
@sarahwalther @mamachanko I assume this change also applies to 1.5.4, 1.6.1 and 1.6.2?

mamachanko:
> @sarahwalther @mamachanko I assume this change also applies to 1.5.4, 1.6.1 and 1.6.2?

Yes! ty

PR #2822, opened 2023-08-10
PDY-1684 Change the verbiage for new configs under cnrs.contour.*

jsanin-vmw:

No need to cherry-pick

### DIFF
--- a/release-notes.hbs.md
+++ b/release-notes.hbs.md
@@ -31,12 +31,14 @@ This release includes the following changes, listed by component and area.
 
 #### <a id='1-7-0-cnrs'></a> v1.7.0 features: Cloud Native Runtimes
 
-- **New config option `contour.default_tls_secret`**: VMware is deprecating the current `default_tls_secret` config option.
-  This new option has the same meaning as `default_tls_secret`. Both config options will be supported for some releases.
+- **New config option `cnrs.contour.default_tls_secret`**: This option has the same meaning as `cnrs.default_tls_secret`.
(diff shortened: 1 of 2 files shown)


jsanin-vmw:
@KauzClay if you can take a look. I addressed some of your comments from previous PR

release-notes.hbs.md
```diff
@@ -31,12 +31,14 @@ This release includes the following changes, listed by component and area.
...
 #### <a id='1-7-0-cnrs'></a> v1.7.0 features: Cloud Native Runtimes
 
-- **New config option `contour.default_tls_secret`**: We are deprecating the current `default_tls_secret` config option.
-  This new option has the same meaning as `default_tls_secret`. Both config options are going to be supported for some releases.
-
-- **New config options `contour.[internal/external].namespace`**: We are deprecating the current `ingress.[internal/external].namespace` config options.
-  This new options have the same meaning as `ingress.[internal/external].namespace`. Both config options are going to be
-  supported for some releases.
+- **New config option `contour.default_tls_secret`**: This option has the same meaning as `default_tls_secret`.
+  `default_tls_secret` has been deprecated in this release and will be removed in CNRs 2.7.0. In the meantime both options
```
KauzClay, line 35: question: I wonder if we should also mention the tentative TAP version this will ship in, since this is TAP docs?

For our own CNR docs, we don't need to, but maybe here we should just for clarity?

jsanin-vmw: good point. We should talk about TAP versions at this level. And in the same line we should be talking about `cnrs.*` properties as well.
I did a little change, let me know what you think

KauzClay: it looks good to me, thanks!

PR #2891, opened 2023-09-06
Services Toolkit: Changes to claim mechanics and resulting RBAC

hoegaarden:

This needs to go into
- main / 1.7


hoegaarden:
Once we ship this in older TAP releases, this needs to be backported to
- 1.6
- 1.5

PR #2877, opened 2023-08-30
Add service registry docs

maly7:
# Which other branches do you want a technical writer to cherry-pick this PR to (if any)?

None, these docs are for TAP 1.7.0

PR #2907, opened 2023-09-08
improve instructions for service registry resource claim

maly7:

this change is only for main/1.7.9


maly7:
@RichardJJG this PR is to fix the issue you mentioned in #2877

RichardJJG:
Thanks, @maly7!

PR #2914, opened 2023-09-12
Launch IntelliJ from CLI please

kdvolder:

This really affects all versions of TAP so should be cherry-picked to as many versions as 
we are still reasonably maintaining docs for. (I.e. in theory: all branches... in practice: as many as you can/want to manage :-)

### DIFF
--- a/intellij-extension/getting-started.hbs.md
+++ b/intellij-extension/getting-started.hbs.md
@@ -18,8 +18,9 @@ This is especially true on Mac OS, because apps launched from Mac OS GUI will re
 restricted set of environment variables compared to a terminal.
 
 This has a tendency to cause problems with cluster authentication for Tanzu Developer Tools. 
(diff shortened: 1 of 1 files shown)


RichardJJG:
Thanks, @kdvolder, will there be a PR flagging the relevant known issue, too, in the various release notes? This problem (it being bad to open IntelliJ through an OS GUI) sounds like a known issue to warn about.

kdvolder:
> relevant known issue
I think there already is something about it in known issues. But PM decided that we needed to do something to elevate its visibility because people keep hitting it.

See here: https://github.com/pivotal/docs-tap/blob/main/intellij-extension/troubleshooting.hbs.md 
Search for the title "Tanzu Panel empty when...".

RichardJJG:
Thanks, @kdvolder, I can see how that suffices.

PR #2944, opened 2023-09-22
Use Package Supply Chain for Ingress

Petahhh:
* suggests for readers of the server workload that Ingress is available out of the box with the package supply chain

* 1-6-3


snneji:
Hi @Petahhh, I assume this will also apply to 1.6.4 too, right?

PR #2958, opened 2023-09-27
Installing IDE extension in airgapped env

kdvolder:

The instructions are not specific to a TAP version. But I suggest adding these notes to Tap 1.7 and Tap 1.6 docs.


kdvolder:
Note: I've tagged the accelerator team in case they wanted to contribute something specific for 'accelerator' IDE exentsions specifically. 

The instructions are fairly generic and the details about Tanzu Developer Tools extension can be seen as a kind of 'concrete example' only. So depending on what they want to do about adding details... you can delete the TODO comment, or it can be replaced with additional concrete information.

kdvolder:
I will add something about accelerator today, so maybe hold-off on merging this until after that.

kdvolder:
Added the details about accelerator, okay to merge now from my perspective.

PR #2966, opened 2023-09-28
Add a pointer to Tap Sandbox from Vscode/IntelliJ about sections

kdvolder:

Probably just Tap 1.7 and later? 

However... this is really not specific to a version of TAP. 

The sandbox environment will most likely be kept up to date so it is always on a 'recent' version of TAP, which means that going back in time and adding a note like this into older versions of the docs makes little sense (you won't get an old version of TAP on the sandbox... and in the future once 1.8 is released you then won't even get a 1.7 version of sandbox anymore soon thereafter.


RichardJJG:
Thanks for that explanation, @kdvolder -- I'll keep this all in `main`.

PR #2971, opened 2023-10-03
docs: 1.7 upgrade docs for store

Manifaust:

This is for 1.7.

The existing docs for upgrading SCST - Store only have instructions for troubleshooting. With 1.7, we introduce the AMR component. In a full profile, the AMR does not need any user configuration. However, in a multicluster TAP, the user does need to do some configuration.

In a previous PR, we had already updated the SCST - Store multicluster docs page. In this MR, we refer to those instructions by expanding the Upgrading doc and linking to the multicluster doc. After that, we also add a link to the TAP upgrading doc.


shaileshpandey:
It looks good to me

PR #2972, opened 2023-10-04
Fix typo in get command for IVS

elfotografo007:
This change is intended for 1.7 and 1.6 branches.

### DIFF
--- a/scst-scan/verify-app-scanning.hbs.md
+++ b/scst-scan/verify-app-scanning.hbs.md
@@ -31,7 +31,7 @@ To verify that you can scan an image using your ImageVulnerabilityScan:
 1. Child resources are created. View the child PipelineRun, TaskRuns, and pods:
 
       ```console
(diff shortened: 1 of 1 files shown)


xtreme-conor-nosal:
As the command is titled: 
> Child resources are created. View the child PipelineRun, TaskRuns, and pods:

I think the typo is actually the first comma, and the intended command is:
`kubectl get -l imagevulnerabilityscan pipelinerun,taskrun,pod -n DEV-NAMESPACE`

This fetches the pr/tr/pods that have an `imagevulnerabilityscan` label applied, so that other prs (e.g. from test stages) are excluded.

RichardJJG:
I've converted to draft mode to prevent accidental merging. When this is definitely ready to merge, please take it out of draft mode.

elfotografo007:
Right, I didn't read the context. I'll update the command now.

PR #2995, opened 2023-10-13
Document AA - AMR integration

ericbottard:
Fix https://github.com/pivotal/acc-release/issues/613

This applies to 1.7 only (main)

### DIFF
--- a/application-accelerator/provenance.hbs.md
+++ b/application-accelerator/provenance.hbs.md
@@ -1,6 +1,10 @@
-# Use the Provenance transform in Application Accelerator
+# Tracking Provenance with Application Accelerator
 
-This topic tells you about the Application Accelerator `Provenance` transform in Tanzu Application Platform (commonly known as TAP).
+Application Accelerator provides several features to track
+the lifecycle of generated projects. This topic is known as "provenance".
+
+## Provenance Transform
+This section tells you about the Application Accelerator `Provenance` transform in Tanzu Application Platform (commonly known as TAP).
 
 The `Provenance` transform is a special transform used to generate a file that
 provides details of the accelerator engine transform.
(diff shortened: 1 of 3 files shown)


application-accelerator/provenance.hbs.md
```diff
@@ -12,4 +16,20 @@ The `Provenance` transform provides traceability and visibility into the generat
...
\ No newline at end of file
+For more information on the structure of the file and how to enable application bootstrapping provenance, see [Provenance transform](creating-accelerators/transforms/provenance.hbs.md).
+
+## Integration with AMR
+Application Accelerator also closely integrates with the [Artifact Metadata Repository](../scst-store/amr/overview.hbs.md) (AMR).
+
+Every time an application is generated with an accelerator, an event is sent to the AMR store that
+roughly captures the same set of information as described in the paragraph above.
+
+The relevant AMR data models are [AppAcceleratorRun](../scst-store/amr/data-model-and-concepts.hbs.md#appacceleratorruns) (one for each invocation, also includes version information
```
kcoriordan, line 27: what does 'one for each invocation' mean here?
If you could put it in sentence structure here, I can update?

application-accelerator/provenance.hbs.md
```diff
@@ -12,4 +16,20 @@ The `Provenance` transform provides traceability and visibility into the generat
...
+about which accelerator was used) and [AppAcceleratorFragments](../scst-store/amr/data-model-and-concepts.hbs.md#appacceleratorfragments) (zero to N, tracks which fragments were used
```
kcoriordan, line 28: What does 'zero to N' mean here?
Can you put it in a sentence, it seems like it lacks context here?

application-accelerator/provenance.hbs.md
ericbottard: I'm not sure I understand the question. It means there is one `AppAcceleratorRun` for each invocation of an accelerator (ie "a run")

application-accelerator/provenance.hbs.md
ericbottard: zero to N means there is between 0 and some number N instances of those, N being the number of fragments used by the accelerator. There is one instance of AppAcceleratorFragment for each (named) fragment used by the accelerator being run.

Hope that helps

PR #3000, opened 2023-10-16
[Read for merge] Add note to say storage class should not have per-node limit

rxl7906:
This PR is missing the correction of the workspace.storageSize default value being 2Gi instead of 100mi which is in this [PR](https://github.com/pivotal/docs-tap/pull/3011/files) 


a-peek4:
@rxl7906 Let me know when you're ready to take this out of draft state. Thank you!

rxl7906:
@a-peek4 Just a heads up, the vulnerability scanning team is going to have draft PRs so that our team members can properly review the PR before merging. But I will let you know which PRs are ready for merge/review!

chaomonica:
Suggestion on the Note:

**Note**: If the StorageClass you select does not have a node limit but uses the node storage (e.g. hostpath), the nodes must have large enough disks. For example, if a scan creates a 2Gi volume on a hostpath type storage class, "2Gi * number of AMR images / number of nodes" would indicate how much storage this cluster would need.

PR #3024, opened 2023-10-20
Services toolkit RBAC edits

ChunyiLyu:
### Which other branches do you want a technical writer to cherry-pick this PR to (if any)?

Main branch please. This only applies to TAP `1.7`


teddyking (commented):
Personally I would revert the changes from line 61-85 in the authorize-claim-provisioner-classes.hbs.md doc as IMO it doesn't add much and actually I think might be conflating issues.

However the rest of the changes LGTM.

PR #3033, opened 2023-10-23
refactored the configurator concepts page

benjaminleesmith:


RichardJJG:
Thanks for the improvements, @benjaminleesmith!

I'll warn now, though, that I'll be changing all the "TDP" instances back to the full name. We can only use brand-name abbreviations that Legal and Marketing have approved.

CC @andreizimin

PR #3036, opened 2023-10-24
VSEE-1168 | Updates to Scan 2.0 Installation page

chaomonica:
- main
- all 1.6.x and 1.7.x branches


rxl7906 (approved):
LGTM

a-peek4:
Hi @chaomonica Please ping me when you are ready to take this out of draft state as the doc freeze is on Nov 3

scst-scan/install-app-scanning.hbs.md
```diff
@@ -116,13 +116,23 @@ To install SCST - Scan 2.0:
 
 The following section describes how to configure service accounts and registry credentials. SCST - Scan 2.0 requires the following access:
 
-  - Read access to the registry containing the Tanzu Application Platform bundles. This is the registry from the [Relocate images to a registry](../install-online/profile.hbs.md#relocate-images-to-a-registry) step or `registry.tanzu.vmware.com`.
-  - Read access to the registry containing the image to scan, if scanning a private image
-  - Write access to the registry to which results are published
+| Registry | Permission | Service Account | Example |
```
ryan-a-baker, line 119: Based on my experience - you are not required to do this if you are just using app-scanning as part of the supply chains right?  I think we should probably put a note at this header that this is optional if you are integrating Scan 2.0 as part of supply chains.  I think the only time you would need to worry about this would be if you are creating standalone IVS templates, right?

scst-scan/install-app-scanning.hbs.md
```diff
@@ -116,13 +116,23 @@ To install SCST - Scan 2.0:
...
+
 
-1. Create a secret `scanning-tap-component-read-creds` with read access to the registry containing the Tanzu Application Platform bundles. This pulls the SCST - Scan 2.0 images.
+1. Create a secret `scanning-tap-component-read-creds` with read access to the registry containing the Tanzu Application Platform bundles. This pulls the SCST - Scan 2.0 images. If you previously relocated the Tanzu Application Platform bundles to your own registry, you can also place your vulnerability scanner image in this registry.
 
-    >**Important** If you followed the directions for [Install Tanzu Application Platform](../install-intro.hbs.md), skip this step and use the `tap-registry` secret with your service account.
+    >**Important** If you followed the directions for [Install Tanzu Application Platform](../install-intro.hbs.md) and did not relocate images, skip this step and use the `tap-registry` secret with your service account. If you did relocate the images from VMware Tanzu Network registry to your own container image registry, skip this step and use the `registry-credentials` secret with your service account.
```
ryan-a-baker, line 135: If we do the above, comment, we probably could remove this disclaimer entirely.

scst-scan/install-app-scanning.hbs.md
ryan-a-baker: I guess from reading below - another scenario where you would want to do this is if the registry that the scan image and target image is different than the images that have the tap bundles and target for workload images that are created by TBS.

scst-scan/install-app-scanning.hbs.md
```diff
@@ -194,12 +220,17 @@ The following section describes how to configure service accounts and registry c
...
+
+8. Apply the service account to your developer namespace by running:
+   ```console
+   kubectl apply -f publisher-sa.yaml
+   ```
 
 ## <a id="registry-retention-policy"></a> (Optional) Set up your registry retention policy
 
-If the registry specified to push scan results to support retention policies, you can configure the registry to delete old scan results automatically depending on your archival requirements. Scan result artifacts accumulate over time and with recurring scanning, artifacts can quickly consume storage space.
+If the registry specified to push scan results to supports retention policies, you can configure the registry to delete old scan results automatically depending on your archival requirements. Scan result artifacts accumulate over time and with recurring scanning, artifacts can quickly consume storage space.
```
ryan-a-baker, line 233:
```suggestion
Although the scan artifacts are ingested in to the Tanzu Application Platform metadata store and information like packages and vulnerabilities parsed are stored, only a pointer to the the original SBOM location stored.  The original SBOM generated by the scan is not preserved within the metadata store.  VMware recommends these original architacts are kept according to your organizations archival requirements.  

Although not recommended, if the registry specified to push scan results to supports retention policies, you can configure the registry to delete old scan results automatically depending on your archival requirements. Scan result artifacts accumulate over time and can quickly consume storage space.
```
Couple thoughts here - we shouldn't recommend people to delete this as they have pointers stored to them within AMR, and it might be important to the customer to keep the original SBOM.  This will get more important as we move to signed sbom's/attestations.

Also, we shouldn't mention recurring scanning as it's not part of the product yet.

ryan-a-baker (commented):
Great job @chaomonica!  I had a couple comments/suggestions, but overall.  Once comments are reviewed/changed then I'll approve.

ryan-a-baker (approved):
That works for me.  We'll see how it goes, I wondered above even moving this to the bottom of the page and out of the normal flow, but I think we can do that when we put it in to profile based installs.

PR #3037, opened 2023-10-24
Document how to uninstall AWS Services

gcapizzi:

None. This only applies to TAP 1.7.

### DIFF
--- a/aws-services/uninstall-aws-services.hbs.md
+++ b/aws-services/uninstall-aws-services.hbs.md
@@ -1,0 +1,144 @@
+# Uninstall AWS Services
+
+This topic tells you how to uninstall the AWS Services package from the Tanzu Application Platform (commonly
+known as TAP) package repository.
+
+## <a id="prepare-infra"></a> Step 1: Prepare your infrastructure
+
+Before uninstalling the package, you will have to decide what to do with the existing resources you have
+claimed.
+
+* In case you want to keep a resource:
+
+  1. Find the Composite Resource associated to your claim:
+
+    ```console
+    kubectl get classclaim <CLASS-CLAIM-NAME> -n <CLASS-CLAIM-NAMESPACE> -ojsonpath="{.status.provisionedResourceRef}"
+    ```
+
+    For example, for a PostgreSQL claim the output will look like this:
+
+    ```json
+    {"apiVersion":"aws.database.tanzu.vmware.com/v1alpha1","kind":"XPostgreSQLInstance","name":"rds-1-r4mgc"}
+    ```
+
+  1. Find the `Instance` associated to the Composite Resource:
+
+      ```console
+      kubectl get <XR-API> <XR-NAME> -ojsonpath="{.spec.resourceRefs}"
+      ```
+
+      For the example above:
+
+      ```console
+      kubectl get xpostgresqlinstance.aws.database.tanzu.vmware.com rds-2-r4mgc -ojsonpath="{.spec.resourceRefs}"
+      ```
+
+      Which will output:
+
+      ```json
+      [{"apiVersion":"rds.aws.upbound.io/v1beta1","kind":"Instance","name":"rds-2-r4mgc-zc69h"}]
+      ```
+
+  1. Edit the `Instance`:
+
+      ```console
+      kubectl edit instance.rds.aws.upbound.io <INSTANCE-NAME>
+      ```
+
+      For the example above:
+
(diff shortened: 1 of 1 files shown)


aws-services/uninstall-aws-services.hbs.md
```diff
@@ -0,0 +1,104 @@
...
+      kubectl get instances.rds.aws.upbound.io
+      ```
+
+  1. Edit the resource:
+
+      ```console
+      kubectl edit instances.rds.aws.upbound.io <RESOURCE-NAME>
+      ```
+
+  1. Change the resource `deletionPolicy` to `Orphan`:
```
ChunyiLyu, line 25: Nitpick, perhaps could mentioned the original value of `deletionPolicy` here, and phrase the sentence like Change the resource `deletionPolicy` from `VALUE` to `Orphan`:

aws-services/uninstall-aws-services.hbs.md
```diff
@@ -0,0 +1,104 @@
...
+
+* In case you want to get rid of a resource:
+
+  1. Delete the corresponding claim:
+
+      ```console
+      tanzu service class-claim delete <CLAIM-NAME>
+      ```
+
+  1. Wait for the resource to disappear in the AWS console.
```
ChunyiLyu, line 45: If users don't delete the class claim and proceed to step 2. What will happen? Just wondering if we should add a sentence to describe that scenario or not.

aws-services/uninstall-aws-services.hbs.md
```diff
@@ -0,0 +1,104 @@
...
+If you don't perform this step, the `ProviderConfig` and its related CRD will hang around in the cluster and
+prevent you from reinstalling the package in the future.
+
+1. Edit the `ProviderConfig`:
+    
+    ```console
+    kubectl edit providerconfig.aws.upbound.io <PROVIDER-CONFIG-NAME>
+    ```
+
+1. Take not of the `Secret` referenced by `spec.credentials`.
```
ChunyiLyu, line 87: `Take not`?

aws-services/uninstall-aws-services.hbs.md
```diff
@@ -0,0 +1,104 @@
...
+1. Delete the `metadata.finalizers` entry.
```
ChunyiLyu, line 88: I think you will need to set `metadata.finalizers` to an empty list, not just delete it. If that's the case, lets be more precise.

aws-services/uninstall-aws-services.hbs.md
gcapizzi: I like it - done!

aws-services/uninstall-aws-services.hbs.md
gcapizzi: If you neither delete the claim nor set the resource to `Orphan`, resources will still survive in AWS but more stuff will be left behind in the cluster (what I saw was the `Instance` + its corresponding CRD).

The reason I don't want to recommend this is that this seems like a lucky accident: the controller gets uninstalled _before_ the `Instance` can be finalised, but I'm not sure how much of a coincidence this is. In the unfortunate event of the `Instance` being deleted before the controller, there's a chance that resources on AWS would be deleted (because of `deletionPolicy: Delete`). I've never seen this in my tests, but I can't explain why.

aws-services/uninstall-aws-services.hbs.md
gcapizzi: "Take note" 😅 fixed!

aws-services/uninstall-aws-services.hbs.md
gcapizzi: I've always just deleted the whole entry in my tests and it's worked 🤷

aws-services/uninstall-aws-services.hbs.md
ChunyiLyu: If you haven't already, perhaps it's worth mentioning something along the line of "if you do not delete class claims before uninstalling, it could leave your TAP and AWS in a dirty state that prevents reinstalling the package..." What do you think?

gcapizzi: Yeah makes sense. I have something similar for the `ProviderConfig` already.

ChunyiLyu (approved):

ChunyiLyu (approved):

PR #3040, opened 2023-10-24
Remove some inconsistencies and redundancies in Azure tap values

ryankilroy:

These values were tested on TAP 1.6.3

### DIFF
--- a/install-azure/profile.hbs.md
+++ b/install-azure/profile.hbs.md
@@ -22,14 +22,6 @@ To relocate images from the VMware Tanzu Network registry to the ACR registry:
 1. Set up environment variables for installation use by running:
 
     ```console
-    export AZURE_SP_APP_ID=MY-AZURE-APP-ID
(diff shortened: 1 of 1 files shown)


install-azure/profile.hbs.md
```diff
@@ -295,15 +282,14 @@ accelerator:
...
-  domain_name: tap.com
 EOF
 ```
 
 Where:
 
-- `tap.com` is a placeholder that must be replaced with a subdomain for the host name that you point at the `tanzu-shared-ingress`
-service's External IP address.
+- `YOUR_DOMAIN` is the subdomain for the host name that you point at the `tanzu-shared-ingress` service's External IP address.
+    > **Note** The `VERSION` and `TAG` numbers differ from the earlier example if you are on
```
YugangW, line 291: Hi @ryankilroy It seems like this sentence is incomplete?

PR #3044, opened 2023-10-25
Update ClusterImageTemplate docs

rxl7906:
Differences between catalog repo [ClusterImageTemplate Trivy](https://gitlab.eng.vmware.com/supply-chain-choreographer/catalog/-/blob/main/src/ootb-templates/bundle/config/image-vulnerability-scan-trivy.yaml) and this ClusterImageTemplate Trivy sample in the Scan 2.0 docs:
* There is `registry.server` and `registry.repository` here but not in the catalog ClusterImageTemplate Trivy because it appears that in the ootb-templates there is ytt happening to populate the registry.server and registry.repository fields but we need to make sure they are populated here if a user wants to create their own ClusterImageTemplate sample.
* The `label_exclusions` in [catalog](https://gitlab.eng.vmware.com/supply-chain-choreographer/catalog/-/blob/main/src/ootb-templates/bundle/config/image-vulnerability-scan-trivy.yaml#L53) are not included in this sample because in 1.7 they defined these values in this [values.yaml ](https://gitlab.eng.vmware.com/supply-chain-choreographer/catalog/-/blob/main/src/ootb-templates/bundle/values.yaml#L7). However if we copied this, we'd get an error trying to fetch this field `label_propagation_exclusions` because it doesn't exist at the time of applying this yaml. 

### DIFF
--- a/scst-scan/clusterimagetemplates.hbs.md
+++ b/scst-scan/clusterimagetemplates.hbs.md
@@ -173,7 +173,7 @@ ClusterImageTemplate Notes:
             publisher: #@ data.values.params.image_scanning_service_account_publisher
           steps:
           - name: trivy-generate-report
-            image: TRIVY-SCANNER-IMAGE # input the location of your Trivy scanner image
+            image: TRIVY-SCANNER-IMAGE
             env:
             - name: TRIVY_DB_REPOSITORY
               value: #@ data.values.params.trivy_db_repository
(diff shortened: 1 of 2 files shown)


a-peek4:
Hi @rxl7906 Please ping me when you are ready to take this out of draft state as the doc freeze is on Nov 3

danataonline (approved):

chaomonica (approved):

rxl7906:
@a-peek4 this has been reviewed and is ready for merge. Thanks!

PR #3053, opened 2023-10-27
Removed Using local-path for publishing acc

jgeethuvmware:


jgeethuvmware:
Make the same changes for TAP 1.6 and any branch after that.

RichardJJG:
> Make the same changes for TAP 1.6 and any branch after that.

Do you mean __before__ that?

jgeethuvmware:
The CLI was deprecated starting TAP 1.6. So need to remove this section from documentation in all the releases starting TAP 1.6 and after that.

PR #3056, opened 2023-10-30
Add MySQL to the AWS Services docs

gcapizzi:

None, this is TAP 1.7.0 only.


aws-services/reference/package-values.hbs.md
```diff
@@ -28,3 +28,20 @@ The following table lists configuration that applies to the `postgresql` service
...
+| mysql.enabled                                    | `false`               | boolean | Enables the MySQL service class                                                                                                                                                                                                                                                   |
+| mysql.infrastructure.security_groups             | _n/a_                 | array   | Security groups for your MySQL databases to belong to                                                                                                                                                                                                                             |
+| mysql.infrastructure.subnet_group.name           | `""`                  | string  | Subnet group for your MySQL databases                                                                                                                                                                                                                                             |
```
ChunyiLyu, line 40: I think it's worth adding description for `security_groups` and `subnet_group` so we are not leaving them blank.

aws-services/reference/supported-services.hbs.md
```diff
@@ -2,7 +2,9 @@
 
 This topic lists the services that are supported by the AWS Services package.
 
-## <a id="postgresql"></a> RDS PostgreSQL
+In Tanzu Application Platform v1.7, the AWS Services package only supports the following AWS services:
```
ChunyiLyu, line 5: Nitpick I would not say `only` here. "the AWS Services package supports the following AWS services" should be sufficient.

aws-services/reference/supported-topologies.hbs.md
```diff
@@ -41,16 +41,28 @@ For instructions for these tasks, see the
...
+    ```yaml
+    mysql:
+      enabled: true
+      region: "REGION"
+      infrastructure:
+        subnet_group:
+          name: "SUBNET-GROUP-NAME"
+        security_groups:
+          - id: "SECURITY-GROUP-ID"
+    ```
```
ChunyiLyu, line 65: I suppose that if one of them, either MySQL or Postgres, is disabled earlier at installation, you don't need to provide values for it?

aws-services/reference/package-values.hbs.md
```diff
@@ -28,3 +28,20 @@ The following table lists configuration that applies to the `postgresql` service
...
+| mysql.instance_configuration.publicly_accessible | `false`               | boolean | Controls whether your instances are publicly accessible                                                                                                                                                                                                                                |
+| mysql.instance_configuration.skip_final_snapshot | `false`               | boolean | Determines whether a final snapshot is created before the instance is deleted. If you specify true, no snapshot is created. If you specify false, a snapshot called final-snapshot-INSTANCE-NAME is created before the instance is deleted.                                            |
+| mysql.provider_config_ref.name                   | `default`             | string  | Name of your ProviderConfig for the `mysql` service                                                                                                                                                                                                                               |
```
ChunyiLyu, line 46: Do we have both postgres and mysql provider config name as `default`? Doesn't that run into conflict?

aws-services/reference/package-values.hbs.md
gcapizzi: They have one 🤔

aws-services/reference/package-values.hbs.md
gcapizzi: Given this is provided by the user, there's no issue - they'll just both use the same `ProviderConfig`. I think `default` gets used anyway as it's also Crossplane's default, but it's the best value we can use for our own default too 🤷

aws-services/reference/supported-services.hbs.md
gcapizzi: I agree - fixed!

aws-services/reference/supported-topologies.hbs.md
gcapizzi: they're not two separate phases - you provide values at installation, and that's how you enable services

aws-services/reference/package-values.hbs.md
ChunyiLyu: My bad. Misread the column

ChunyiLyu (approved):

PR #3060, opened 2023-10-30
tdp default plugins

Braundo:
Added TDP default plugins in `template_variables.yaml` and then referenced in relevant docs files

### DIFF
--- a/tap-gui/configurator/building.hbs.md
+++ b/tap-gui/configurator/building.hbs.md
@@ -47,40 +47,40 @@ To prepare your Configurator configuration file:
 
 1. Create a new file called `tdp-config.yaml` by using the following template:
 
-    ```yaml
-    app:
-      plugins:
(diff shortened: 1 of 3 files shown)


RichardJJG:
Thanks, @Braundo. The 4-space indentation for the YAML was on purpose, by the way, and I've reset that. YAML, within a numbered step, must be indented 4 spaces to render properly because of a DocWorks bug.

RichardJJG:
The bullets in the variable don't render properly.

image

I tried in HTML to no avail:

image

I've changed these back to regular text. I've created a DocWorks [feature request](https://jira.eng.vmware.com/browse/CDF-21467?src=confmacro) to make rendering bullets via HBS variables possible.

Braundo:
@RichardJJG - do you know how the values currently present in the template_variables.yaml file are used in the docs? I tried to mimic that functionality, but evidently missed something.

PR #3063, opened 2023-10-31
Update troubleshoot for gitops delete-kubernetes-auth.sh

kumaritanushree:


RichardJJG:
@kumaritanushree, which other branches do you want a technical writer to cherry-pick this PR to (if any)?

neil-hickey (approved):
I think this makes the most sense - given we don't have release notes or anything. The funny part is that this applies only to one single patch release. So would someone find this here? Maybe? Probably the best place for now. LGTM

kumaritanushree:
@RichardJJG these changes are done only for 1.7.0 and error will be seen only for this. Hence it should be part of 1.7.0 onwards only.

PR #3069, opened 2023-10-31
Add more details about TDP entity and API curation

shrutiyer:
- Fix the links into variables

Question for the tech writer: Are you able to test the links? We want to variable-ize all the links to keep it in one place to quickly check and update them as new versions come in.

### DIFF
--- a/api-auto-registration/api-curation.hbs.md
+++ b/api-auto-registration/api-curation.hbs.md
@@ -130,7 +130,7 @@ To retrieve curated API specifications:
   api-auto-registration-controller   api-auto-registration.tap.domain  api-auto-registration-cert   valid    Valid HTTPProxy
   ```
 
-1. To get all of the curated API specifications with curl, use FQDN with an http scheme:
+1. To get all the curated API specifications with curl, use FQDN with an http scheme:
 
   ```console
   curl http(s)://AAR-CONTROLLER-FQDN/openapi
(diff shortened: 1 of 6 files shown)


shrutiyer:
@karinepires @csterwa @visomar @tiagofernandez check out the MR for any additions

csterwa (approved):
I think it is a bit strange to put the Backstage links under AAR in variables. Should they be put somewhere else?

visomar (approved):

RichardJJG:
2 reasons not to put all those links in variables:

1. It's harder to spot when they fall out of date within VS Code. Link-checking extensions won't work.
2. It's difficult to see where the links go at all -- the engineer/tech writer will have to keep checking `template_variables.yaml`.

shrutiyer:
@RichardJJG so for tech writers, is it recommended to not do this? Sounds like it makes things difficult for y'all? Our team was trying to do this because we don't how many links to update (for example for all the SCG links) when we have released a new version of that other product. Is there a recommendation to not have stale but still functional links?

RichardJJG:
> @RichardJJG so for tech writers, is it recommended to not do this? Sounds like it makes things difficult for y'all? Our team was trying to do this because we don't how many links to update (for example for all the SCG links) when we have released a new version of that other product. Is there a recommendation to not have stale but still functional links?

I sympathize with wanting to keep links fresh for minimal effort. There is a [2-year-old ticket](https://jira.eng.vmware.com/browse/CDF-18843) for being able to create links that always redirect to the latest version.

Is the 2.1 version number the main problem here? If so, you could just keep that number in a variable. That's what we did with the [TAP version variable](https://github.com/pivotal/docs-tap/blob/main/template_variables.yaml#L3).

shrutiyer:
Sounds good. Thanks for the suggestion and I have made the updates to follow the pattern other components are following. Ready for your review @RichardJJG

RichardJJG:
> Sounds good. Thanks for the suggestion and I have made the updates to follow the pattern other components are following. Ready for your review @RichardJJG

Thanks, @shrutiyer -- that does look easier to handle going forward.

shrutiyer:
No other branches as of now. Thanks!

PR #3079, opened 2023-11-02
added backstage <-> tap version table

Braundo:
1.7+ only


benjaminleesmith:
lgtm

PR #3095, opened 2023-11-03
Remove references to RBAC CLI

ryan-a-baker:

As of right now, just the main branch.  However, we will cherry pick this back to 1.4 on release date as I don't want to do it yet because we'll remove the binary from tanzunet on release date.  Will log separate PR's for that.


snneji:
@ryan-a-baker, is the plan to also remove this from the next patch releases for 1.6, 1.5, 1.4?

PR #3110, opened 2023-11-07
Create durable `tanzu cli plugin install --group....` command

heyjcollins:

TAP 1.8+

With Tanzu CLI v1.1.0, we can use the MAJOR.MINOR reference in the `tanzu plugin install --group vmware-tap/default:vX.Y` command and it will always install the latest patch version for the MAJOR.MINOR version specified in the command.
This eliminates the need to update the Tanzu CLI plugin installation command docs every time a new Patch release GA's for a given Minor release line.


kcoriordan:
thanks for PR - I'll also add this variable in following files which also reference the TAP plugin group version.
docs-tap\cli-plugins\tanzu-cli.hbs.md
docs-tap\cli-plugins\apps\getting-started\installation.hbs.md

@heyjcollins

PR #3125, opened 2023-11-13
added apigroup to cluster-view-setup

Braundo:
only for 1.7.0+

[Slack thread here](https://vmware.slack.com/archives/C02D60T1ZDJ/p1699465190998079)


RichardJJG:
Thanks, but we had this change PR'ed to main and I've already cherry-picked it back here.
https://github.com/pivotal/docs-tap/commit/bd4c41ca0609483e72d8c8ecb015aaf2995820af

PR #3130, opened 2023-11-14
revert mitigation for switching envoy from daemonset to deployment

KauzClay:

### DIFF
--- a/contour/how-to-guides/configuring-envoy.hbs.md
+++ b/contour/how-to-guides/configuring-envoy.hbs.md
@@ -27,4 +27,4 @@ contour:
 
 ## <a id="daemonset"></a> Configure the Envoy pods as a `Deployment`
 
-Currently, we do not have steps to make this switch without downtime. Look out for improvements in future versions!
+Currently, we do not have steps to make this switch without downtime.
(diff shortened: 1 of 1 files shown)


contour/how-to-guides/configuring-envoy.hbs.md
```diff
@@ -25,110 +25,6 @@ contour:
...
-   tanzu package installed update tap -p tap.tanzu.vmware.com -v $TAP_VERSION  --values-file tap-values.yaml -n tap-install
-   ```
-
-5. (Optional) Bring the affinity rules back
-
-   If you wish restore the affinity rules on the Envoy Deployment, simply remove the `contour-remove-affinity` overlay from your `tap-values.yaml` file, and update again.
-
-   This will cause the Deployment to roll out new pods, so be mindful of how many replicas you have. Fewer replicas will increase the likelyhood of a disruption in service as the pods are rolled out.
\ No newline at end of file
+Currently, we do not have steps to make this switch without downtime.
```
xtreme-sameer-vohra, line 30: Suggestion: drop this line because Clay has determined a way of doing this.

PR #3150, opened 2023-11-23
chore: add a default step template into the pipeline

garethjevans:
For TAP 1.7


mdc81:
@snneji @YugangW this was supposed to go into 1.7.1. Looks like its in main only as of now.

PR #3152, opened 2023-11-24
chore: document changes required for Kaniko to work with PSA restricted

garethjevans:
For 1.7.1


scc/dockerfile-based-builds.hbs.md
```diff
@@ -127,9 +127,25 @@ use of the SCC, OpenShift accepts the pods created to run Kaniko to build
 the container images.
 
 
+> **Note** Such restrictions are due to well-known limitations in how Kaniko
```
kcoriordan, line 130: Is it correct to have the note repeated here - it's already on line 145?
It seems out of context here as the limitation being described here is because of the default unprivileged
SecurityContextConstraints (SCC)?

PR #3192, opened 2023-12-11
VSEE-1287 | added a note to specifically call out TKGs 1.26+ as restricted cluster

bsoroushian:
where affinity assistance should be specifically be disabled.

please cherry pick to `main`.

### DIFF
--- a/scst-scan/install-app-scanning.hbs.md
+++ b/scst-scan/install-app-scanning.hbs.md
@@ -40,7 +40,7 @@ tekton_pipelines:
     disable_affinity_assistant: "true"
 ```
 
-Having affinity assistance `enabled` specifically on clusters with restricted pod context like TKGs 1.26+ where Pod Scurity Admission Pod Security Standards is set to be restricted by default, can cause a deadlock where affinity pod cannot be scheduled and will prevent scanning go ahead.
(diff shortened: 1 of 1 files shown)


scst-scan/install-app-scanning.hbs.md
```diff
@@ -40,6 +40,8 @@ tekton_pipelines:
     disable_affinity_assistant: "true"
 ```
 
+Having affinity assistance `enabled` specifically on clusters with restricted pod context like TKGs 1.26+ where Pod Scurity Admission Pod Security Standards is set to be restricted by default, can cause a deadlock where affinity pod cannot be scheduled and will prevent scanning go ahead.
```
elfotografo007, line 43: Here is a suggestion to make it a little bit more clear:

```suggestion
Having the Affinity Assistant feature `enabled` on clusters where the Kubernetes Pod Security Admission is enforcing Pod Security Standards to be restricted by default (like TKGs 1.26+), can cause a deadlock where the affinity pod cannot be scheduled and it will prevent scanning go ahead.
```

ryan-a-baker: With Andres's changes I'm good with this.

PR #3197, opened 2023-12-20
Remove net-istio config options

andrew-su:
* 1-7-0
* 1-7-1
* 1-7-2
* 1-7-3

---
The package no longer bundles net-istio. Cleaned out references to configuration options specific to net-istio. Additionally these specific values will always cause a deploy to fail.


andrew-su:
Closing this for now, we will revisit after we begin the deprecation for the two options.

andrew-su:
As those options were never valid in the first place, I propose we reopen and apply these docs changes.

KauzClay:
looks good to me 👍

PR #3200, opened 2023-12-27
Added label "part-of" in workload sample of example1 docker workload deployment steps

sendjainabhi:
Added label "part-of" in workload sample of example1 docker workload deployment steps


YugangW:
Thanks @sendjainabhi. Do you want to cherry-pick this to 1.7.x as well?

scc/dockerfile-based-builds.hbs.md
```diff
@@ -47,6 +47,7 @@ Dockerfile by passing the `dockerfile` parameter:
 $ tanzu apps workload create foo \
```
djschny, line 47: This should ideally be prefixed with something different like `my-` so that way it's a different value that `part-of`.

PR #3203, opened 2024-01-03
Fix indenting

joostvdg:
The indentation of `ingress` was off.


RichardJJG:
Thanks, @joostvdg. This indentation is needed in all earlier docs versions, too, right?

joostvdg:
@RichardJJG I haven't checked other versions. If they have the same structure as this, then yes.

PR #3206, opened 2024-01-04
better clarity on example values for `dockerfile-based-builds.hbs.md`

djschny:
The idea here is to make it more clear which values are example values and to distinguish between example values and what they mean instead using the same value for all examples.

### DIFF
--- a/scc/dockerfile-based-builds.hbs.md
+++ b/scc/dockerfile-based-builds.hbs.md
@@ -32,20 +32,20 @@ without running Docker inside a container.
     <td>
       The list of flags to pass directly to kaniko, such as providing arguments to a build.
     </td>
-    <td><pre>- --build-arg=FOO=BAR</pre></td>
+    <td><pre>- --build-arg=MY_KEY=MY_VALUE</pre></td>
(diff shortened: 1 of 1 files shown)


RichardJJG:
@djschny, a writer should cherry-pick this back as far as possible, right?

djschny:
> @djschny, a writer should cherry-pick this back as far as possible, right?

I don't know. That's up to you and your internal procedures.

RichardJJG:
@djschny, a writer shouldn't cherry-pick backwards if SCC is so deceptively different in some earlier version that this placeholder change would be misleading. Your response suggests that's not the case, though, so a writer will cherry-pick backwards as far as the code examples exist.

djschny:
Once again, I don't believe that is the responsibility of the contributor. The PR is for the main branch. If there were differences/deltas on older versions, that should be separate PRs to there. Cherry picking is dangerous and clunky.

RichardJJG:
@djschny, there should, indeed, be separate PRs if the PR content must be modified to apply to earlier versions of the product.

PR #3245, opened 2024-01-23
Use a simple unauthenticated example to start the walkthrough

Manifaust:
The existing example uses an authenticated example which requires an access token. We should start with an unauthenticated endpoint to ease the reader.

Merge into 1.7, 1.6, 1.5, and 1.4.

### DIFF
--- a/scst-store/api-walkthrough.hbs.md
+++ b/scst-store/api-walkthrough.hbs.md
@@ -6,22 +6,42 @@ This topic includes an example API call that you can use with Supply Chain Secur
 
 The following procedure explains how to use CURL to POST an image report.
 
-1. In your terminal, switch to the kubectl context or kubeconfig to target the TAP _View_ cluster.
+1. In the terminal, switch to the kubectl context or kubeconfig to target the TAP _View_ cluster.
 
-2. Retrieve the `metadata-store-read-write-client` access token.
-    See [Retrieve access tokens](retrieve-access-tokens.hbs.md). Run:
-
-    ```console
-    export METADATA_STORE_ACCESS_TOKEN=$(kubectl get secrets metadata-store-read-write-client -n metadata-store -o jsonpath="{.data.token}" | base64 -d)
-    ```
-
-3. Retrieve the CA Certificate and store it locally. Run:
+2. Retrieve the CA certificate and store it locally. Run:
 
     ```console
     kubectl get secret ingress-cert -n metadata-store -o json | jq -r '.data."ca.crt"' | base64 -d > /tmp/ca.crt
     ```
 
-4. Run the Curl POST command. This assumes that you have deployed TAP with ingress enabled. We discuss in a [different section](#without-ingress) what to do if ingress is not enabled.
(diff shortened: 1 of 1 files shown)


kcoriordan:
@Manifaust 
Hi Tony
For this topic [API walkthrough for Supply Chain Security Tools - Store](https://docs-staging.vmware.com/en/draft/VMware-Tanzu-Application-Platform/1.7.4/tap/scst-store-api-walkthrough.html), 
Step 3 of the Without Ingress sections says - 'Continue following the instructions from the previous section.' From what step forward should we indicate here? Is it from step 3 forward?

Also I would like to change the structure of this topic to tabs

Manifaust:
> From what step forward should we indicate here? Is it from step 3 forward?

They should start from step 1 of the previous section. If you think step 1 is redundant (duplicating step 1 from the bottom section), then we can start from step 2. But I think step 1 serves as a good reminder of where the customer should be situated.
 
> Also I would like to change the structure of this topic to tabs

If you think it would be easier to read, then I think we can try tabs.

Manifaust:
I think your question raises an issue I had missed, so thank you. The steps for non-ingress users and port forwarding is actually more involved than this. In this particular page being edited in the PR, those steps had been cut improperly through last year as we moved the emphasis to the ingress users.

The non-ingress use case is so small that I think we should probably not spend too much doc space on it. Fortunately, in a previous commit we had already created docs that we can point to. I have forgotten about these pages but we should link users to them.

- scst-store/use-node-port.hbs.md
- scst-store/use-load-balancer.hbs.md

Instead of creating a new section called "Without ingress", I think I should have further and better de-emphasized that secondary use case by just creating a small **Note** box that pointed users to that page.

That will clean up the doc, but potentially raise the cognitive burden for users with that user case because they have to jump back and forth between docs and also know where to resume. But my argument is that those users are so small (none?) and probably just technically advanced Broadcom-internal testers that they can handle the lack of hand-holding. For the majority use case, it will make the doc more readable.

What do you think?

kcoriordan:
OK, thanks @Manifaust makes sense.
- I've added a note now and done some other style edits. The style for variables is caps with dashes, so for INGRESS-DOMAIN, I removed the <>
In step 5, in this section `-H "Authorization: Bearer ${METADATA_STORE_ACCESS_TOKEN}" \`
Should we indicate here that `METADATA_STORE_ACCESS_TOKEN` is the value from the previous step?
Is the METADATA_STORE_ACCESS_TOKEN variable style consistent with the INGRESS-DOMAIN variable style?
Take a look here: 
https://docs-staging.vmware.com/en/draft/VMware-Tanzu-Application-Platform/1.7.4/tap/scst-store-api-walkthrough.html

Manifaust:
I think your style changes still work yes.

As for `METADATA_STORE_ACCESS_TOKEN`. I don't think you need to indicate that the value is from the previous, as a user with the prerequisite skills should know this already. And we should keep with the underscore style for `METADATA_STORE_ACCESS_TOKEN`, since it is following variable convention from BASH.

PR #3275, opened 2024-02-06
Remove auth references from DORA plugin docs

davewalter:
- As of TAP 1.9, the plugin no longer requires authentication to display DORA metrics.

#### Which other branches do you want a technical writer to cherry-pick this PR to (if any)?

None. This is a change for TAP 1.9 only.


snneji:
Hi @davewalter, since this is for 1.9, could you put it in draft mode until 1.8 is released. `main` is currently 1.8 and I don't want us to accidentally merge it into that branch. Thanks!

PR #3276, opened 2024-02-06
Docs for Backstage RBAC Plugin

nandy-vmware:


snneji:
Hi @nandy-vmware, is this for any versions other than 1.8?

nandy-vmware:
Hi @snneji, no. This is only for 1.8.

PR #3299, opened 2024-02-09
Adds CNR upgrade bug troubleshooting

Petahhh:

* `1-7-4`
* `1-6-8`

### DIFF
--- a/cloud-native-runtimes/how-to-guides/troubleshooting.hbs.md
+++ b/cloud-native-runtimes/how-to-guides/troubleshooting.hbs.md
@@ -2,6 +2,30 @@
 
 This topic tells you how to troubleshoot Cloud Native Runtimes, commonly known as CNRs, installation, or configuration.
 
+## <a id='updates-fail'></a> Updates fail with annotation is immutable
+
+### Symptom
+
+After creating a web workload in TAP 1.6.3 or below, then upgrading beyond TAP 1.6.3, and attempting to update your web workload, you will see the following error
+
+```
+API server says: admission webhook "validation.webhook.serving.knative.dev" denied the request: validation failed: annotation value is immutable: metadata.annotations.serving.knative.dev/creator (reason: BadRequest)
+```
+
+### Explanation
+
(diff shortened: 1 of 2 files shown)


cloud-native-runtimes/how-to-guides/troubleshooting.hbs.md
```diff
@@ -2,6 +2,30 @@
...
+
+After creating a web workload in TAP 1.6.3 or below, then upgrading beyond TAP 1.6.3, and attempting to update your web workload, you will see the following error
+
+```
+API server says: admission webhook "validation.webhook.serving.knative.dev" denied the request: validation failed: annotation value is immutable: metadata.annotations.serving.knative.dev/creator (reason: BadRequest)
+```
+
+### Explanation
+
+Knative adds annotations to Knative Services. Kapp Controller, which is the orchestrator underneath workloads, does validation that created objects are not modified by others. This conflict between Kapp Controller and Knative is a known and expected behaviour but is mitigated by a configuration passed into Kapp Controller. This configuration specifies that the annotations Knative adds are ok and thus should not fail the validation. This configuration is added dynamically in the delivery supply chain.
```
KauzClay, line 17: nit:

you refer to it as "a configuration passed into Kapp Controller". Personally, I don't think you need to allude to it like this. Rather, I think you can just call it a "kapp Config" ([docs](https://carvel.dev/kapp/docs/v0.45.0/config/#overview)).

and then "this configuration.." could become "the kapp Config" or "the kapp Config file".

cloud-native-runtimes/how-to-guides/troubleshooting.hbs.md
```diff
@@ -2,6 +2,30 @@
...
+
+As of TAP 1.6.4, the configuration moved from the delivery supply chain to the build supply chain. Thus, when a web workload is being updated, the delivery supply chain is missing this configuration resulting in the validation error. Though the configuration exists on 1.6.4 but in a different part of the supply chains, existing deliverables are not rebuilt to include the config.
+
+### Solution
+
+We are currently in the process of providing further support to resolve this issue. Below is an immediate work around. We will publish an additional workaround and will resolve this issue for TAP 1.6.9, 1.7.5 and 1.8.1 in which simply upgrading will prevent and resolve this issue. A second workaround will be published shortly ensuring no downtime.
```
KauzClay, line 23: nit:

I think we're saying too much here.

In my opinion, we can just say something like:

```
Below is an immediate workaround. Alternate workarounds, as well as a proper resolution, are in development.
```

KauzClay (commented):
just a few nits for thought

PR #3304, opened 2024-02-12
Update recurring scan docs to fix output and add update db step

kkavitha:
Needed for tap 1.8.0+


ryan-a-baker:
Nevermind.  The PR compare view made this look really weird.  This looks good to me.

PR #3351, opened 2024-02-21
Update amr-observer size recommendations

derwei:
Please cherry pick this into the 1.7 and 1.8 branches. Thank you!


gabbbar:
@derwei  This change is not seen in 1.7 release, why to patch in 1.7 ?

derwei:
@gabbbar I was thinking it would be good to cherry-pick to the 1.7 branch just in case. But if you feel that is unnecessary, I'm ok with leaving this in 1.8 only

PR #3357, opened 2024-02-22
fix suply chain install to not use yq

laidbackware:
This PR affects TAP 1.8 only.

The manual install instructions for Supply Chains had a dependency on `yq` but the same outcome can be achieved with kubectl and the jsonpath output method.


squeedee:
LGTM!

PR #3358, opened 2024-02-22
Add note about the scan step

kkavitha:

This MR is only for TAP 1.8.0


kcoriordan:
https://docs-staging.vmware.com/en/draft/VMware-Tanzu-Application-Platform/1.8/tap/scst-scan-recurring-scanning.html

PR #3374, opened 2024-02-23
Update scale example for amr observer to reflect change in best practice table

derwei:
Please add this change to the same branches as https://github.com/pivotal/docs-tap/pull/3351#issuecomment-1960783000

In that PR, I mentioned 1.7 and 1.8 - but a comment made after the merge, it might not be necessary to have it in the 1.7 branch.


derwei:
Oops, saw that this was fixed

PR #3375, opened 2024-02-23
Add known issue for TANZUSC-6133

derwei:
TAP 1.8 only please and thank you!


mdc81:
@derwei I have not tried the Scan 2.0 with where non default issuer is used. Do you if that works successfully?

derwei:
@mdc81 We have added in another PR to remove that recommendation. https://github.com/pivotal/docs-tap/pull/3386

PR #3448, opened 2024-03-12
Fix work around and add more detail to KI

derwei:
Please cherry pick this to the 1.8.x branch as well.
Thanks!

### DIFF
--- a/release-notes.hbs.md
+++ b/release-notes.hbs.md
@@ -117,11 +117,11 @@ This release has the following known issues, listed by component and area.
 
 #### <a id='1-9-0-scst-policy-ki'></a> v1.9.0 Known issues: Supply Chain Security Tools - Policy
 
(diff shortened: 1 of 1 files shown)


release-notes.hbs.md
```diff
@@ -117,11 +117,11 @@ This release has the following known issues, listed by component and area.
 
 #### <a id='1-9-0-scst-policy-ki'></a> v1.9.0 Known issues: Supply Chain Security Tools - Policy
 
-- Supply Chain Security Tools - Policy is defaulting to TUF enabled. This can cause the package to not reconcile correctly if the default TUF mirrors are not reachable.
-To work around this, explicitly configure policy controller in TAP values to not enable TUF
```
kcoriordan, line 121: should the first sentence say '.. is defaulting to TUF disabled ..' rather than enabled?

derwei: The wording "defaulting to TUF enabled" is correct - but this is incorrect behaviour, so we want to call it out with a work around.

PR #3449, opened 2024-03-14
There can still be issues with managed resources if Synced=True

Samze:
1.8.x and above


Samze:
@snneji I don't think this was backported to 1.8.x. Are you able to do that?

PR #3469, opened 2024-03-25
Condensed install log for CC and TSC

rashedkvm:
- `main` only


RichardJJG:
Hi, @rashedkvm, we use ```console instead of ```shell because we expect that, whenever we get linter support for the docs, ```console is the safer choice for consistent syntax highlighting across the docs.

PR #3472, opened 2024-03-27
Add note about breaking change in upgrade docs

derwei:
Not sure if this is the best place to put this, but we'd like to call out this breaking change in the upgrade docs as well as we anticipate that users will run into this issue.

This is just for the 1.9.x branch (I'm not sure if this should carry over to future branches... maybe since users can potentially upgrade from 1.7, 1.8 -> 1.10)


derwei:
Please ping me on slack if there are any questions/suggestions around this =)

mdc81:
> Not sure if this is the best place to put this, but we'd like to call out this breaking change in the upgrade docs as well as we anticipate that users will run into this issue.
> 
> This is just for the 1.9.x branch (I'm not sure if this should carry over to future branches... maybe since users can potentially upgrade from 1.7, 1.8 -> 1.10)

Yes, its needed for doc branches 1.9.x and above.

PR #3496, opened 2024-04-03
Aws schema engine version

Samze:

TAP 1.10


snneji:
@Samze since this is for 1.10, I'm going to convert to draft for now. We'll merge into main after 1.9 is released.

Samze:
Undrafting now that 1.9 is released

PR #3497, opened 2024-04-03
[TANZUSC-6352] Add release note for known issue around image_scanning_cli.image

rxl7906:
Summary:
* Add release note to provide workaround for the known issue of workload failing at the image scanning step when `image_scanner_template_name: image-vulnerability-scan-trivy` is specified without `image_scanning_cli.image`

1.8.2

### DIFF
--- a/release-notes.hbs.md
+++ b/release-notes.hbs.md
@@ -163,6 +163,21 @@ enable TUF:
     tuf_enabled: true
   ```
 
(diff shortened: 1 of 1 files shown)


snneji:
@rxl7906 does this issue also affect 1.8.0 and 1.8.1? If so, we will need to copy it to the known issues section for those versions as well.

mdc81:
> @rxl7906 does this issue also affect 1.8.0 and 1.8.1? If so, we will need to copy it to the known issues sections for those versions as well.

yes. It does affect 1.8.0 and 1.8.1 as well.

PR #3514, opened 2024-04-05
Adds http-proxy to the TDP TOC

mstergianis:
This is also the thing that defines the side nav right? Really I want this to show up in the side nav so that folks can discover the doc.

just this one, this is a new feature with 1.9


snneji:
We received your PR after the Release Lock date. We are currently working on release tasks, and we will review your PR as soon as we are able. Unfortunately, this might mean that your PR content will not be published until after the release. Going forward, to ensure that your PR changes are published on release day, please submit such PRs no later than 23:59 UTC-8 on the [Release Lock date](https://confluence.eng.vmware.com/pages/viewpage.action?pageId=789919969).

RichardJJG:
I did this work outside of the PR

PR #3519, opened 2024-04-08
Add Bitnami 1.10 breaking change

Samze:
1.10


Samze:
Keeping in draft as its for TAP 1.10 as per this comment https://github.com/pivotal/docs-tap/pull/3496#issuecomment-2035122883 from @snneji

